PR #20, opened 2023-02-24
I propose fix for https://github.com/bwhitman/amy/issues/19 by

octetta:
chosing a specific libsoundio backend in the Linux case.

### DIFF
--- a/src/libsoundio-audio.c
+++ b/src/libsoundio-audio.c
@@ -122,10 +122,15 @@ amy_err_t soundio_init() {
         fprintf(stderr, "out of memory\n");
         return 1;
     }
+#ifdef __linux__
+    int err;
+    err = soundio_connect_backend(soundio, SoundIoBackendPulseAudio);
+#else
     int err = soundio_connect(soundio);
+#endif
     if (err) {
         fprintf(stderr, "Unable to connect to backend: %s\n", soundio_strerror(err));
-        return 1;
+        exit(1);
     }
 
     soundio_flush_events(soundio);


octetta:
> Great. Tested on Ubuntu 20, Fedora 36, macOS. All of them work. Thanks so much @ocetta! 
> 
> 

Nice. I also successfully tested this on a RPi3 as well.

I suspect some Linux system won't be happy with me forcing to PulseAudio though.

Do you think there's value in modifying the soundio_init function to allow the backend to be user-selected?

I can work on a PR with my idea if you think it's worth the effort.

bwhitman:
I'm happy to hear from others, but my belief on `amy-example` is that it's really just a quick way to show AMY working on your system (outside of an ESP32.) Anyone building AMY into a product will likely use their own audio library. 

It's awesome that now `amy-example` works on more Linux platforms than it did yesterday, thanks to you! but i don't know if spending a ton of time making it work for all sorts of edge cases like different audio backends is immediately valuable. There's probably even a better thing than `libsoundio` we should be using, that was really just me finding the easiest solution at the time. 

Once AMY is embedded into more projects, we'll probably get a better handle on what we need from a desktop audio layer.

octetta:
> I'm happy to hear from others, but my belief on `amy-example` is that it's really just a quick way to show AMY working on your system (outside of an ESP32.) Anyone building AMY into a product will likely use their own audio library.
> 
> It's awesome that now `amy-example` works on more Linux platforms than it did yesterday, thanks to you! but i don't know if spending a ton of time making it work for all sorts of edge cases like different audio backends is immediately valuable. There's probably even a better thing than `libsoundio` we should be using, that was really just me finding the easiest solution at the time.
> 
> Once AMY is embedded into more projects, we'll probably get a better handle on what we need from a desktop audio layer.

This makes total sense and I appreciate the pragmatism. My selfish goals are to use AMY inside the Elixir environment, kind of like a 21st century POKEY or SID chip... indeed, AMY caught my attention because of it's name and homage. I saw press about Atari's work years ago and always wanted that silicon. Thanks for building something like it for me!

PR #22, opened 2023-03-02
Show use of miniaudio via new command that accepts entered AMY ASCII commands.

octetta:
As per https://github.com/bwhitman/amy/issues/21 this PR shows using a different audio library.

Has been primarily tested on Fedora 37, but also macOS Ventura, and Raspbian.

```
make -f Makefile.message
./amy-message
v0f440l1
v1f441l1
v2f442l1
```


bwhitman:
The first thing I noticed (other than it _just working_, which was great, with no library installs) is that the `-c` parameter does not work as expected. I use `-c` to address individual outputs of a multi-channel sound card for things like Alles running on a computer. 

I'll take a moment to try to shim that back in, but need to see how miniaudio does its channel handling.

bwhitman:
Oh, ugh, i created a new PR by accident , but I think I got channels in there

bwhitman:
OK, we're back. Take a look at what I did and lmk if it still works for you? You can try `-c 0` or `-c 1` and it should only play on that channel

octetta:
> OK, we're back. Take a look at what I did and lmk if it still works for you? You can try `-c 0` or `-c 1` and it should only play on that channel

Oh. I totally removed the channel part. Thanks for putting this back in. It works.
The only thing remaining is I forgot to put `amy-message` in `.gitignore`.

octetta (commented):
The channel change looks perfect (sorry for totally removing what you had before).

bwhitman:
Great. I merged this in. Thanks so much for all your help @octetta . I'll work on porting this over to Alles and Tulip later too, it'll simply things a bit there.

PR #27, opened 2023-03-02
Add libraries to `Makefile` so AMY compiles on RPi/ARM/32-bit Linux

octetta:
As per https://github.com/bwhitman/amy/issues/25
Tested on RPi3.


bwhitman (commented):
libatomic is not installed by default (at least) on macOS, so we need to detect RPI here if it's required there

PR #32, opened 2023-03-08
Create function to return sample rate amy was compiled with.

octetta:
Trying to address https://github.com/bwhitman/amy/issues/28


davay42 (commented):
Looks quite nice! I've fixed tuning issues just by hardcoding 48000 there. This approach seems much more reliable!

octetta:
Cool. I'm stacking commits, hopefully this doesn't cause troubles. I have another PR related to the `amy-message` that I snuck in earlier.

davay42:
No problem, just looking around. 👀 I've discovered AMY just recently and am very interested in using it in my webapps. But need to settle the most performant and reliable link between JS and WASM. I wonder how `src/www/amy.js` file gets generated?

octetta:
`amy.js` (and `amy.wasm` as well) is generated by the Emscripten compiler `emcc`, as a result of `make web`.

Not to distract from AMY, which I quite like, I've played around with a version of PureData implemented as a JS Web Audio library as well.

I suspect you and I are more focused on the web and desktop variants at the moment... hopefully Brian and DAn will keep us true to their vision. :100:

davay42:
I find AMY low level enough to be able to build many playful music things on top without all the wiring. WebPd looks promising, but it can be too much for my use cases.

With AMY's oscillator syntax we can invent new patching systems for any implementation. Just need to set a solid foundation for that. I still struggle to make it play sounds with Vue 3 and mobile (vanilla JS works fine). And also we really need an AudioWorklet version of the JS-WASM connection to be able to use the synth with all the power of ToneJS.

bwhitman:
Thank you @octetta !

PR #47, opened 2023-07-08
pcm_samples: Fix 'wrap-around' click in sample 11 (808-CNGLO2-D)

dpwe:
`pcm_samples_*.h` contain banks of short PCM samples, represented as int16s.  It turns out these samples contained exactly one value of -32768, which, judging by the waveform, should have been +32768 but got wrapped around in the conversion to int16.  The result was a distinct 'click' at the start of PCM sample 11.  This change replaces this one value (in both files) to +32767.

Thanks to the very sharp ears of @drepetto for finding this problem!


dpwe:
See https://colab.research.google.com/drive/1WxGh9THgLek3aILQVB5DCkOIifItk6A0#scrollTo=nFjmUPtRStyE

PR #90, opened 2023-12-26
Remove the __weak__ attrs, just let users set AMY_CORES and PCM_SIZE at compile time. Everything else you'd edit `amy_config.h` for.

bwhitman:


bwhitman:
The `_weak_` stuff was causing more problems (in python, and Arduino) than it was worth

bwhitman:
Now it's like `amy_start(cores, reverb, chorus)`. None of those are evaluated a-rate so shouldn't slow down things

examples/AMY_Test_ESP32/AMY_Test_ESP32.ino
dpwe, line 1: If this is supposed to be didactic, maybe name the parameters in comments, e.g. 
```
amy.begin(/* cores= */ 1, /* reverb= */ 0, /* chorus= */ 0);
```

dpwe: Looks great.  I agree that making these parameters settable at run-time has minimal downside.

PR #118, opened 2024-05-29
fixing midi note not resetting logfreq

bwhitman:


bwhitman:
Dan confirms this is not the right way to do this

PR #120, opened 2024-06-08
Replace TARGET_* with ControlCoeffs and ENVELOPE_*.

dpwe:
This removes the `TARGET_*` defines in `amy.h`, and the Amy wire protocol commands that use the `TARGET` bit sets (`bp?_target`, `mod_target`).  We don't need these because the functions that are affected by the EGs and the mod input are now set by coefficient values in the ControlCoefficient vectors for `amp`, `freq`, `filter_freq`, `duty`, and `pan`.

There were a few places (in fm.py and examples.c) where we were still using the `TARGET` pattern; these have been converted.

I also undid [fixing midi note not clearing logfreq_coefs for PCM](https://github.com/bwhitman/amy/commit/145283ab61b5cd8289d2e41114229fd42f9ef507) as has long been my secret goal.


bwhitman:
Thank you dan!!!!!!!

PR #122, opened 2024-06-08
external hook

bwhitman:


dpwe:
Looks good.

* I would pass num_samples to the external hook.  *Of course* it's always `AMY_BLOCK_SIZE`, but this external hook may be ignorant about Amy internals.  It just makes sense as a signature IMO.
* `SAMPLES` is not the greatest type to be exporting - it's Amy-internal, and somewhat nonstandard (sign bit, 8 integer bits, 23 fractional bits).  As it stands, the external hook is going to have to be pretty well-informed about Amy internals.  I guess that's OK.  The alternative would be to convert to int16 (like the waveform output) or float, but I can see why you don't want to do that.
* Why not have the function return a bool saying whether it consumed the samples?  Less clumsy than a pointer to a return value.

bwhitman:
OK on (1). on (2) i agree but not sure i want to do all that work per block. (3) sure, but i am not sure how to set up a function pointer that returns a value. i can ask the internet

bwhitman:
Fixed up for (1) and (3). Gonna merge

PR #129, opened 2024-06-22
pcm.c,amy.c: Fix looping in PCM samples.

dpwe:
The core problem was using `PCM_INDEX_FRAC_BITS` when it needed to be `PCM_INDEX_BITS`.

But in the process of trying to figure out how it worked, I think I found a bug with `hold_and_modify()` continually re-asserting the `msynth[osc].feedback` field which was being used by `pcm_note_off()` to indicate when a looping sample should be allowed to play out.

Added a new test, `TestPcmLoop`, to verify continued correct operation.


bwhitman:
👷‍♂️

Issue #140, opened 2024-08-06
`partials.py` seems out of sync with main AMY


bwhitman:
From discord user `mathias`:

>Guess I found an error in partials.py:

```
import partials
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mb/py_test/Amy/partials.py", line 40, in <module>
    def sequence(filename, max_len_s = 10, amp_floor=-30, hop_time=0.04, max_oscs=amy.OSCS, freq_res = 10, freq_drift=20, analysis_window = 100):
                                                                                  ^^^^^^^^
AttributeError: module 'amy' has no attribute 'OSCS'
```
 
>In amy.py the max_osc-Identifyer (now?) is called MAX_OSCS  After I changed the name of the referenced constant importing partials to use Loris as described in the docs works. Of course I am not entirely sure what's the intended solution - the change could be done to OSCS instead of MAX_OSCS in amy.py instead, of course. But to adjust partials.py accordingly appeared to be more logical to me...
>There is another error in partials.py:
```
"bp0_target":amy.TARGET_AMP+amy.TARGET_LINEAR,
            "bp1_target":amy.TARGET_FREQ+amy.TARGET_LINEAR,
```
>I deleted those lines because TARGET_AMP and TARGET_FREQ are not defined in amy.py
>Plus during playback there is an error from amy.py, that bp0_target and bp1_target are no valid messages
>(had to close the shell and only can report from the top of my head - sorry!)
>Now the partials can be played back, but I did not hear anything and the next time there are errors:
```
partials.play(s, amp_ratio=2)
386.0
>>> partials.play(s, amp_ratio=2)
AMY queue is full
AMY queue is full
```

bwhitman:
Working on this now.

I'm fixing #9 by forcing `--prefix` to be a local directory, which helps with an install (especially using virtualenv or conda or whatever)

Fixed up `AMY_OSCS`

The `bp0` and `bp1` have to be rewritten and using ControlCoefs, so we remove `TARGET_AMP` and `TARGET_FREQ`

PR #151, opened 2024-08-14
Get `partials.py` and `partials.c` up to date with latest AMY, including Loris decomposition playback

bwhitman:
Fixes for #140 and #9


bwhitman:
thanks @dpwe !!!

PR #166, opened 2024-08-18
Set unset CtrlCoefs set through C struct API to 0

### LINKED ISSUE #163: `mod_source` not modifying correctly
In C, this:

```c
struct event a = amy_default_event();
a.osc = 1;
a.wave = SINE;
a.freq_coefs[COEF_CONST] = .25;
a.amp_coefs[COEF_CONST] = 1;
amy_add_event(a);

struct event e = amy_default_event();
e.osc = 0;
e.wave = PULSE;
e.freq_coefs[COEF_CONST] = 440;
e.mod_source = 1;
e.duty_coefs[COEF_MOD] = 1;
e.velocity = 1;
amy_add_event(e);
```

Should make a 0.25Hz duty cycle LFO on a 440Hz pulse wave. But it plays a constant pulse wave instead. 

Written in Python, it "just works":

```python
amy.reset()
amy.send(osc=1,wave=amy.SINE,freq=0.25,amp=1)
amy.send(osc=0,wave=amy.PULSE,mod_source=1, freq=440, duty="0,0,0,0,0,1",vel=1)
```

Something is up with how we fill in `X_coefs` in Python that makes assumptions we're not getting in C, I _think_

bwhitman:
This fixes #163 for me. This will do precisely what the string parsing function will do, but for direct C `struct event` API users. If they only set one of the 7 CtrlCoefs, the other 6 will be filled in as 0. This gets us equivalence to the Python examples/API.

### DIFF
--- a/src/amy.c
+++ b/src/amy.c
@@ -397,6 +397,7 @@ void amy_add_event(struct event e) {
 void amy_add_event_internal(struct event e, uint16_t base_osc) {
     AMY_PROFILE_START(AMY_ADD_EVENT)
     struct delta d;
+    uint8_t setflag;
 
     // Synth defaults if not set, these are required for the delta struct
     if(AMY_IS_UNSET(e.osc)) { e.osc = 0; } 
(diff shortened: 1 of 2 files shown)


src/amy.c
```diff
@@ -425,32 +426,63 @@ void amy_add_event_internal(struct event e, uint16_t base_osc) {
...
+        for (int i = 0; i < NUM_COMBO_COEFS; ++i) {
+            if(AMY_IS_UNSET(e.freq_coefs[i])) e.freq_coefs[i] = 0;
             float freq_coef = e.freq_coefs[i];
             // Const freq coef is in Hz, rest are linear.
             if (i == COEF_CONST) freq_coef = logfreq_of_freq(freq_coef);
             d.param=FREQ + i; d.data = *(uint32_t *)&freq_coef; add_delta_to_queue(d);
         }
     }
 
-    if(AMY_IS_SET(e.filter_freq_coefs[COEF_CONST])) { float filter_logfreq = logfreq_of_freq(e.filter_freq_coefs[COEF_CONST]); d.param=FILTER_FREQ; d.data = *(uint32_t *)&filter_logfreq; add_delta_to_queue(d); }
```
bwhitman, line 440: This was weird and bears a closer looking at -- we were checking filter_freq_coefs[COEF_CONST] twice, but doing the same thing each time. I have it just looking at it once now

dpwe: Yes, you're right.  `[filter_]freq_coefs[COEF_CONST]` is a special case as it is converted from Hz to logfreq units, unlike the remaining parameters.  But I'd put the special case in twice.  I like your resolution.

However, I disagree with clearing unset values to 0 (the remainder of this PR).  I think the C interface should support setting individual components of the `_coefs` vectors one at a time.  We pack them into a vector for the Python interface for brevity and to save on wire-protocol symbols, but there are plenty of use-cases for changing them individually (this is exactly what you want to happen when you move the VCF Freq slider on your MIDI controller).

dpwe (commented):
I don't want to do this, C users should be able to change individual control coefs without affecting the rest. The reason Python doesn't do this is because the syntax of specifying "just the 3rd coefficient" is awkward, and also because it neatly meant the existing tutorial kept working. But that tutorial hides the gotcha of the preset coef values that need to be reset. I'd prefer changing the tutorial, not limiting the flexibility of the C interface.

bwhitman:
>I don't want to do this, C users should be able to change individual control coefs without affecting the rest.

I'm a bit confused -- that's precisely what the PR is attempting to do. It wasn't possible before, without manually setting each coef to something. But with this, now you can do 

```c
e.amp_coefs[COEF_MOD] = 1;
```

And it'll work. You can also do


```c
e.amp_coefs[COEF_MOD] = 1;
e.amp_coefs[COEF_VEL] = 0.5;
```

And it'll work, just like it does in Python. All that's happening is when it gets a single update to any of the coefs, it checks if any others are not set, and sets them to 0. That's precisely what python is doing. What am I missing?

bwhitman:
We have subsumed this one with a better PR

PR #202, opened 2024-08-31
Adds Portamento

dpwe:
A new parameter `portamento=<time_ms>` (`m` in wire protocol) sets a time-constant over which note changes (without intervening note-offs) are applied.


dpwe:
This is not yet integrated into the synth patches.  I suspect it will work directly with DX7, since I think the osc 0 pitch drives the whole note.  But for Juno, it will require some intermediation, since all three of the (pulse, sawtooth, subosc) oscs need to have their portamentos locked to the same value to make sure the whole note shifts as one.

PR #205, opened 2024-09-07
Add echo effect.

dpwe:
Adds a new mix-level time effect, echo.  Provide wire protocol ('M') for `amy.send(echo='LEVEL,DELAY_MS,MAX_DELAY_MS,FEEDBACK,FILTER_COEF')`.  Also adds an extra parameter to C function `amy_start()` to optionally enable echo, since it will consume something like 256kB of RAM by default once it is first turned up.


bwhitman:
amazing, thank you dan!!

PR #221, opened 2024-09-24
README.md: Fix the FM example to clear amp[VEL].

dpwe:
The example that plays a couple of FM tone by directly configuring the individual sine oscillators had been broken since we made the change that unspecified ControlCoefficients are left unchanged (rather than being cleared to zero).  Because FM component oscs do not see the velocity of their parent note, it's important to include `amp={'vel': 0}` (to clear the default weight of 1), otherwise the operator oscs will always have zero output, because they see a velocity of zero.

Doing this, I realize that `ratio` is really just a special case of ControlCoefficients.  Instead of `ratio=0.2` (and the associated explanation), that's really just `freq={'const':  52.326, 'note': 1}`, i.e. it follows the note frequency, but relative to 0.2 x the default frequency (of 261.63 Hz).  We could change that.  Would it make things easier to understand?

### DIFF
--- a/README.md
+++ b/README.md
@@ -438,42 +438,43 @@ When building your own algorithm sets, assign a separate oscillator as wave=`ALG
 
 ```python
 amy.reset()
-amy.send(osc=0, wave=amy.SINE, ratio=0.2, amp='0.1,0,0,1', bp0='0,1,1000,0,0,0')
-amy.send(osc=1, wave=amy.SINE, ratio=1, amp=1)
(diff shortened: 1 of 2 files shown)


dpwe:
I realize using the "dict" form of specifying **ControlCoefficients** will weird-out people unfamiliar with Python.  At the same time, it's actually a lot more comprehensible than `'1,0,0,1'` once you learn the syntax.

PR #227, opened 2024-09-29
reset_osc with RESET_TIMEBASE resets time clock

bwhitman:
For _reasons_ (Amy chip, and amy-web) I would like to be able to send AMY a message to reset its time base to 0, so that future calls with time=x start from 0. 

I _personally_ see this as a form of “reset”. We already have the notion that reset_osc set to > AMY_OSCS (120) resets all of the oscs. 

What I posit to you is what if we had an even bigger reset, like reset_osc = 10000 , that also restarted audio (including the audio clock?)


bwhitman:
Also, we remove the `amy.send(time=0)` in `test.py`, because the `amy.restart()` takes care of that already.

dpwe:
These values of reset seem pretty janky. Maybe make it a bitmask? `amy.send(reset=amy.RESET_OSCS | amy.RESET_TIME, ...)`

bwhitman:
Bitmask is a good idea.

bwhitman:
How about this?

dpwe (approved):

PR #228, opened 2024-09-30
Remove clone_osc from AMY

bwhitman:
clone_osc was used early on in juno development, but subsumed by `voices`. It's no longer used anywhere except for two tests, so I rewrote those and removed clone_osc (which was mostly commented out code anyway.)


bwhitman:
This is now subsumed by #231

PR #229, opened 2024-09-30
Moving memorypcm inside AMY

bwhitman:
This puts the memorypcm code inside AMY. 

You can load samples with AMY's `amy.transfer_wav(wav_filename)`, which works on Tulip and normal Python. It encodes the wav file as base64 chunks and transfers them using a new AMY command `z/load_sample`. 

If you want to use the wire protocol outside of Python, you need to find a base64 encoder and follow what `amy.transfer_wav()` is doing. It's simple enough that it should work on most other implementations.

If you're linking AMY directly (and not using wire protocol), you can fill in the allocated sample buffer that is returned after calling `int16_t * pcm_load(uint16_t patch, uint32_t length, uint32_t samplerate, uint8_t midinote, uint32_t loopstart, uint32_t loopend)` yourself.


bwhitman:
A wire protocol to load samples is a nice to have but not required. Developers (e.g. in Tulip) can load in samples using `pcm_load`, and AMY will malloc ram for the sample like we do for reverb/etc. But it would be nice to have a way to load in sample data over the wire protocol. Here's a stab at what that would look like:

`C/load_sample`: 
`load_sample` takes a comma separated list of `patch, samplerate, midinote, loopstart, loopend`. Only `patch` is required. `samplerate` defaults to 22050. `midinote` defaults to 60. `loopstart` and `loopend` default to 0. `patch` must be 1024 and higher. 

After `load_sample` is sent, any further input to AMY is treated as a base64 encoded string over any amount of messages. That base64 string is concatenated and decoded in AMY and loaded into newly allocated sample RAM. The base64 message(s) must end with a `#`, which will be discarded (and is never sent during base64 encodes). After the sample is decoded and loaded, normal AMY messages can follow.

bwhitman:
This is now subsumed by #231

PR #230, opened 2024-10-01
Audio in support (no buses, just a AUDIO_IN0 wave type)

bwhitman:
Start of audio in support for AMY. This lets you set wave=AUDIO_IN0 or AUDIO_IN1 to an oscillator and set its `vel` to nonzero. You can then add effects, filters, etc after. 
The audio subsystem should fill in amy_in_block every block before rendering starts. I'll add this into miniaudio and it currently works using [`amychip`](https://github.com/shorepine/amychip)


bwhitman:
This is all set for now, I think. We can tackle buses (or whatever we want) later. I tested this with `pyamy/libamy`, Tulip Desktop, `amychip`, `amy-example`.

bwhitman:
This is now subsumed by #231

PR #238, opened 2024-10-28
A minor documentation fix, and addition of a missing file in pico's cmakelists

vijaymarupudi:
Note that I have not tested the pico `CMakeLists.txt` change, I used those files to make my own `CMakeLists.txt` for my own project (non-pico), and noticed that the code wouldn't compile if I didn't add the `transfer.c` file.


bwhitman:
Thanks!

PR #239, opened 2024-11-02
Smooth retrigger

dpwe:
Under this change, we add state for each BP_SET that records which segment it was in (`synth[osc].current_seg[bp_set]`), and has a dynamic field for the start value of the current envelope segment (`synth[osc].seg_start_val[bp_set]`).  When `compute_breakpoint_scale` detects that the envelope is in a new segment, it sets `seg_start_val` to the most recently-returned scale value (`last_scale`) and the envelope segment proceeds from there.

This is quite nice in that it subsumes the previous behavior of release segments (which started from wherever you had gotten to) and extends it to the attack segment of retriggered notes.  If you want to always restart your attacks from 0, simply add a `'0,0,'` prefix to your envelope spec.

This was to allow smooth retriggering of waveforms as suggested by @vijaymarupudi on Discord.


vijaymarupudi:
This is great, thank you very much again! I'm learning a lot by reading the code you've written, hopefully I can contribute to amy one day!

While the clicking in the sine waves is fixed, I seem to be hearing clicking when I use square waves, but only with a filter (and I suspect, only when I use note ons?). Just implemented recording functionality in my codebase to provide more data.

Screenshot from 2024-11-02 17-25-33

Here's a link to the `.wav` file if you want it, Github doesn't seem to allow me to upload it directly here: https://gtvault-my.sharepoint.com/personal/vmarupudi3_gatech_edu/_layouts/15/download.aspx?share=EVrEEvvLloJPsxnsiB6kREgB9vtVjejUBDZZrSRO5IEK2A

dpwe:
I moved the filter reset to occur only at the same time as the phase reset.  I think that will fix the discontinuity you saw, and it seems to leave most things unharmed.  There were small changes in several of the tests, I think because some notes may terminate a frame or two earlier, but nothing I could notice by ear.

vijaymarupudi:
I can confirm that the issue is fixed for the monophonic square wave! I can't hear any clicks at all! Thank you so much!

PR #243, opened 2024-11-06
Sequencer in AMY

bwhitman:
This PR adds the sequencer to AMY. We have new `sequence` and `tempo` commands. 

This allows anyone to have a Tulip-style clock sequencer running directly in AMY. 

README.md has the API for it and some examples. 

This gets us closer to fixing up https://github.com/shorepine/tulipcc/issues/340

### DIFF
--- a/src/sequencer.c
+++ b/src/sequencer.c
@@ -144,8 +144,8 @@ static void sequencer_check_and_fill() {
     }
 }
 
-// Desktop: called from a thread
-#ifndef ESP_PLATFORM
+// posix: called from a thread
+#if defined _POSIX_THREADS
 void * run_sequencer(void *vargs) {
     // Loop forever, checking for time and sleeping
     while(1) {
@@ -156,7 +156,7 @@ void * run_sequencer(void *vargs) {
 }
 
 // ESP: do it with hardware timer
-#else
+#elif defined ESP_PLATFORM
 static void sequencer_timer_callback(void* arg) {
     sequencer_check_and_fill();
 }
--- a/src/sequencer.c
+++ b/src/sequencer.c
@@ -1,0 +1,233 @@
+// timer.c 
+// handle timer-like events for music
+// runs in a thread / task
+
+/* 
+AMY style sequencer spec
+for now: fixed ppq (48)
+
+you can pass "sequence" / H as a string,up to 3 values, tick,divider,tag 
+if divider is 4, then it's every 4 ticks (48 ppq at bpm)
+tag is a numerical constant, can be whatever (uint32)
+sequence=",4,123" -- no tick, just divider, tag = 123
+sequence is passed along other messages, but is treated like time=X  -- not played until the sequence fires
+if sequence="59320,,123" then this is an absolute tick in the future and played only once
+you can remove scheduled sequences with sequencer=",,123" -- nothing in the tick section will delete a message if there is a tag
+tick takes precedence over divider if both given
+
+ok, how we deal with it internally:
+
+in the parse_amy_message, we check for sequence, if it's there, we set the event type to EVENT_SEQUENCE and send it to
+sequence_add_event(struct event e, tick, divider, tag) 
+
+sequence_add_event checks / parses the sequence string and fills in a new entry in the seqeunce entry LL
+it also can remove sequence entries if no tick is given and the tag exists 
+
+a thread/timer that checks every 48ppq tick with BPM 
+and then adds it when it's time to play
+it adds it by amy_adding_event with a re-set time variable and status = EVENT_SCHEDULED
+entries that have ticks get deleted after adding
+
+we also call a callback function if set void sequencer_callback(uint32_t tick)
+
+*/
+
+#include "sequencer.h"
+
+// Things that people can change
+float sequencer_bpm = 108; // verified optimal BPM 
+
+
+typedef struct  {
+    struct event e; // parsed event -- we clear out sequence and time here obv
+    uint32_t tag;
+    uint32_t tick; // 0 means not used 
+    uint16_t divider; // 0 means not used 
+} sequence_entry;
+
+// linked list of sequence_entries
+typedef struct sequence_entry_ll_t{
+    sequence_entry *sequence;
+    struct sequence_entry_ll_t *next;
+} sequence_entry_ll_t;
+
+sequence_entry_ll_t * sequence_entry_ll_start;
+
+// Optional sequencer hook that's called every tick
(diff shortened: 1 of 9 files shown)


bwhitman:
(as a side effect, we also cleaned up `parse_int_list` etc with a macro that works with 5 types instead of repeating the code.)

src/sequencer.c
```diff
@@ -0,0 +1,227 @@
+// timer.c
```
dpwe, line 1: File name seems to have changed.

src/sequencer.c
```diff
@@ -0,0 +1,227 @@
...
+
+ok, how we deal with it internally:
+
+in the parse_amy_message, we check for sequence, if it's there, we set the event type to EVENT_SEQUENCE and send it to
+sequence_add_event(struct event e, tick, divider, tag) 
+
+sequence_add_event checks / parses the sequence string and fills in a new entry in the seqeunce entry LL
+it also can remove sequence entries if no tick is given and the tag exists 
+
+a thread/timer that checks every 48ppq tick with BPM
```
dpwe, line 26: This appears to be the start of a new, perhaps older description. Punctuation can help interpret.

src/sequencer.c
```diff
@@ -0,0 +1,227 @@
...
+void parse_tick_and_tag(char * message, uint32_t *tick, uint16_t *divider, uint32_t *tag) {
+    uint32_t vals[3];
+    parse_list_uint32_t(message, vals, 3, 0);
+    *tick = (uint32_t)vals[0]; *divider = (uint16_t)vals[1]; *tag = (uint32_t)vals[2];
+}
+
+uint8_t sequencer_add_event(struct event e, uint32_t tick, uint16_t divider, uint32_t tag) {
+    // add this event to the list of sequencer events in the LL
+    // if the tag already exists - if there's tick/divider, overwrite, if there's no tick / divider, we should remove the entry
+    if(divider != 0 && tick != 0) { divider = 0; } // only use tick if both set
```
dpwe, line 84: Comment is ambiguous. "If tick is set, ignore divider."

src/sequencer.c
```diff
@@ -0,0 +1,227 @@
...
+
+    sequence_entry_ll_t *entry_ll = sequence_entry_ll_start;
+    sequence_entry_ll_t *prev_entry_ll = NULL;
```
dpwe, line 87: Isn't the normal way to do this as parent_p = &root_ptr, then access current as (*parent_ptr)...

src/sequencer.c
```diff
@@ -0,0 +1,227 @@
...
+    uint8_t found = 0;
+    while(entry_ll != NULL) {
+        if(entry_ll->sequence->tag == tag) {
+            found = 1;
+            if(divider == 0 && tick == 0) { // delete 
+               if(prev_entry_ll == NULL) { // start
```
dpwe, line 93: .. because then this logic would subsumed

dpwe (approved):
Lgtm

src/sequencer.c
```diff
@@ -0,0 +1,193 @@
...
+                }
+                free(entry_ll->sequence);
+                free(entry_ll);
+                return 0;
+            } else { //replace 
+                entry_ll->sequence->divider = divider;
+                entry_ll->sequence->tick = tick;
+                entry_ll->sequence->e = e;
+                return 0;
+            }
```
dpwe, line 72: It's impossible to get here, right?  So we don't need found, because any time we set found, we end up either deleting or replacing, and doing an early return.

src/sequencer.c
```diff
@@ -0,0 +1,193 @@
...
+typedef struct  {
+    struct event e; // parsed event -- we clear out sequence and time here obv
+    uint32_t tag;
+    uint32_t tick; // 0 means not used 
+    uint16_t divider; // 0 means not used 
+} sequence_entry;
+
+// linked list of sequence_entries
+typedef struct sequence_entry_ll_t{
+    sequence_entry *sequence;
```
dpwe, line 12: I think if you make this an actual sequence_entry rather than a (sequence_entry *), you'd only have to malloc once for each sequence_entry_ll, and you could remove a level of indirection, i.e. you'd be accessing (*entry_ll_ptr)->sequence.tag etc.

src/sequencer.c
```diff
@@ -0,0 +1,193 @@
...
+    parse_list_uint32_t(message, vals, 3, 0);
+    *tick = (uint32_t)vals[0]; *divider = (uint16_t)vals[1]; *tag = (uint32_t)vals[2];
+}
+
+uint8_t sequencer_add_event(struct event e, uint32_t tick, uint16_t divider, uint32_t tag) {
+    // add this event to the list of sequencer events in the LL
+    // if the tag already exists - if there's tick/divider, overwrite, if there's no tick / divider, we should remove the entry
+    if(divider != 0 && tick != 0) { divider = 0; } // if tick is set ignore divider 
+
+    sequence_entry_ll_t *entry_ll = sequence_entry_ll_start;
```
dpwe, line 52: sequence_entry_ll_t **entry_ll_ptr = &sequence_entry_ll_start;  // Start pointing to the root node.
while ((*entry_ll_ptr) != NULL) {
  if ((*entry_ll_ptr)->sequence->tag == tag) {
    if (divider == 0 && tick == 0) { // delete
      sequence_entry_ll_t doomed = *entry_ll_ptr;
      *entry_ll_ptr = doomed->next;  // close up list.
      free(doomed->sequence);
      free(doomed);
      return 0;
    } else { // replace
      (*entry_ll_ptr)->sequence->divider = divider;
      (*entry_ll_ptr)->sequence->tick = tick;
      (*entry_ll_ptr)->sequence->e = e;
      return 0;
    }
  }
  entry_ll_ptr = &((*entry_ll_ptr)->next);  // Update to point to the `next` field in the preceding list node.
}
// If we got here, we didn't find the tag in the list, so add it at the end.
if(tick == 0 && divider == 0) return 0;   // Ignore non-schedulable event.
if(tick != 0 && tick <= sequencer_tick_count) return 0; // don't schedule things in the past.
(*entry_ll_ptr) = malloc(sizeof(sequence_entry_ll_t));
(*entry_ll_ptr)->sequence = malloc(sizeof(sequence_entry));
(*entry_ll_ptr)->sequence->e = e;
(*entry_ll_ptr)->sequence->tick = tick;
(*entry_ll_ptr)->sequence->divider = divider;
(*entry_ll_ptr)->sequence->tag = tag;
(*entry_ll_ptr)->next = NULL;
return 1;
}

dpwe (commented):
I added comments showing how I think working with **ll_item would work

PR #249, opened 2024-11-19
Add examples with I2S pin defintions

znmeb:
I've added examples with the I2S pin values set for three boards that I have:

Pimoroni PicoVision - Pico 2 W with onboard I2S DAC.
Pimoroni Pico Audio Pack - I2S DAC with both Line and Headphone out. I tested this with a Pico 2.
Waveshare Pico Evaluation Board - I2S DAC with Headphone out (and other I/O hardware). I tested this with a Pico 2.

I have a couple of other boards but these are the only three I'm planning to test in the near future.


znmeb:
Cancelling pull request - I moved these to a different project

PR #254, opened 2024-12-24
Piano partials

bwhitman:
This only changes one part of AMY - `#define MAX_BREAKPOINTS 16` -- the rest are breakpoint and residual files and notebooks. So I think ok to merge ?


dpwe (commented):
I am fine to merge all this stuff.  The JSON files are hardly canonical, but they could be interesting.  I'm really not sure what state piano_heterodyne.ipynb is in, it's mid-WIP but I don't think there's much harm in releasing it.

bwhitman:
Check out the piano patch maker, now in `amy_headers.py` -- test it like

```python
# make test
# python3
>>> import amy; amy.live()
>>> amy.send(load_patch=256, voices='0')
>>> amy.send(voices='0',note=55,vel=1)
```

bwhitman:
thank you dan !!!!

PR #267, opened 2025-01-07
added to allow for two mod sources, mod0 and mod1

koltenpearson:
For the project I am working on I wanted the ability to have 2 LFOs so I went ahead and added support for it.

If this is not desired then no worries, go ahead and close it.

If this is desired then note I only made the minimal number of changes to support my use case. Specifically I did not try to support python, and I imagine it breaks the examples and tests. I figured I would wait and see if this was wanted before going any farther. 

I am willing to add support to the python bindings examples and tests, though I will probably need a little guidance on how you would want the syntax and argument names to be in the interest of backwards compatibility and whatnot.

### DIFF
--- a/src/amy.c
+++ b/src/amy.c
@@ -386,7 +386,8 @@ struct event amy_default_event() {
     AMY_UNSET(e.portamento_ms);
     AMY_UNSET(e.filter_type);
     AMY_UNSET(e.chained_osc);
-    AMY_UNSET(e.mod_source);
+    AMY_UNSET(e.mod0_source);
+    AMY_UNSET(e.mod1_source);
     AMY_UNSET(e.eq_l);
     AMY_UNSET(e.eq_m);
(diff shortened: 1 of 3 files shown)


bwhitman:
Nice! We'll take a look.

dpwe:
Nice work. 

I am curious about the 2 LFO use case - is it for pitch and envelope at separate rates or something like that?

koltenpearson:
Yes being able to change both pitch and amplitude/pwm on the same oscillator is the use case I had in mind. It is more for me to experiment with than achieve a specific sound.

dpwe:
Hi, I'm sorry I never got around to integrating this, I think it's a good idea.  I'm going to get Claude to do it from scratch since it's better and finding all the places that might be implicated, but I want to acknowledge that it's due to your suggestion!

PR #275, opened 2025-01-08
add support for syncing oscillator waveforms

koltenpearson:
This makes it so amy can simulate analog oscillator syncing. It works by specifying on the synced oscillator which oscillator to sync to,  and when computing the waveform, it will use that oscillators frequency to reset its own waveform. This should let it work intuitively for people familiar  with how analog synths implement syncing.

This shouldn't break any existing functionality, but there would still some work to do to add python support, so let me know if this is a desired feature and I can work towards adding the python bindings.

### DIFF
--- a/src/amy.c
+++ b/src/amy.c
@@ -387,6 +387,7 @@ struct event amy_default_event() {
     AMY_UNSET(e.filter_type);
     AMY_UNSET(e.chained_osc);
     AMY_UNSET(e.mod_source);
+    AMY_UNSET(e.sync_source);
     AMY_UNSET(e.eq_l);
     AMY_UNSET(e.eq_m);
     AMY_UNSET(e.eq_h);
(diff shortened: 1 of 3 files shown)


koltenpearson:
My approach here isn't bandlimited. I see there are techniques to do band limited syncing e.g. https://www.cs.cmu.edu/~eli/papers/icmc01-hardsync.pdf, they are a little beyond me at the moment, and I haven't had much luck finding existing implementations. Here is one in scheme? https://ccrma.stanford.edu/~dfl/220b/hw4.htm

dpwe:
I'm not familiar with osc syncing (as you can see, my design goal was emulating a Juno-6 which doesn't have it).  I see the aliasing problems that can arise with summarily resetting the LUT phasor here.  How does this sound?  Is the aliasing problematic?  I don't know what it's supposed to sound like...

I'll also note that wire protocol letters are an extremely scarce resource, so there's a somewhat higher utility threshold before we deploy one of the few remaining.  I'm trying to figure out ways to work around this, for instance by aliasing certain letters (e.g. both `patch_num` and `num_oscillators` are `p`, whose meaning depends on whether `wave=BYO_PARTIALS` is in the message), but I don't see any obvious candidates in this case.

dpwe:
So, apparently the sync input on the [CEM3340](https://electricdruid.net/wp-content/uploads/2020/02/CEM33403345-VCO.pdf) analog VCO chip beloved of 1980 synths is more complicated than simply resetting the phase to zero: https://www.youtube.com/watch?v=b4-emEhVoyY&t=320s

koltenpearson:
Yeah that datasheet mentions that there are many different was to do syncing on that VCO chip. I was hoping for the simplest (in general not necessarily on that chip), which is just setting the phase to 0. But my research has shown that pretty much the only techniques for making band limited syncing work are minblep or polyblep. Which is probably beyond me at the moment.

I can probably live with the aliasing for now for my project, I am not so good a sound designer that I am confident knowing how good or bad it is yet. At some point I need to stop programming and start making music if I want to get  there ; )

znmeb:
FWIW I just discovered this textbook on digital synth algorithms: https://link.springer.com/book/10.1007/978-3-031-14228-4. Pretty much every algorithm I've ever seen is in there. The downside is that it assumes floating point arithmetic, so there would be extra work to incorporate the algorithms into AMY, but the main ones are already in AMY.

PR #284, opened 2025-01-28
Volume code

dpwe:
When converting amy.py:message() to use a lookup dictionary, I miscoded 'volume' to use 'v' in the wire protocol (osc) instead of 'V' (capital).  So attempting to alter volume from MicroPython did .. nothing.


dpwe:
Also includes unrelated updates to piano_heterodyne.ipynb, where I was checking that the UIowa piano samples did in fact contain the note they claimed (all but one, Piano.C1.pp, are correct).

PR #295, opened 2025-02-04
Small updates to interp-partial

dpwe:


dpwe:
This is the PR that I'm using with 6-note polyphony piano on Tulip hardware

PR #298, opened 2025-02-04
Run dpwe's script to generate amy_constants.py on build

bwhitman:


Makefile
```diff
@@ -55,6 +55,7 @@ HEADERS_BUILD := $(filter-out src/patches.h,$(HEADERS))
 PYTHONS = $(wildcard *.py)
 
 src/patches.h: $(PYTHONS) $(HEADERS_BUILD)
+	cat src/amy{,_config}.h  | sed -e 's@^//.*@@' | egrep 'define +[^ ]+ +[.0-9-]+' | sed -e 's/\([.0-9]\)f$$/\1/' | awk '{if ($$3 == "") $$3="True"; print $$2 "=" $$3}' > amy_constants.py
```
dpwe, line 58: There's no need for  the `if ($$3 == "")` awk clause.  That was before I added the egrep to limit it to literal numbers, and was to handle the `#define AMY_FIXEDPT`.    Now it's just `awk '{print $$2 "=" $$3}'`.

dpwe: looks great otherwise

PR #300, opened 2025-02-06
Remove define AMY_SEQUENCER_PPQ from sequencer.h

dpwe:
We just moved it to amy.h


bwhitman:
oops! Thanks for noticing that

PR #315, opened 2025-03-31
Hooks for external audio, EXT0/EXT1

bwhitman:
This allows external audio channels EXT0 and EXT1 to be set outside of amy. It also supports a per-frame callback so that outside programs can get/set the audio buffers.


bwhitman:
`make test` is fine and this doesn't impact anything audio wise, so i'm going to merge this

Issue #302, opened 2025-02-11
PROPOSAL: AMY should handle MIDI in and out


bwhitman:
Since we have AMY handling a real time sequencer, we should also allow it to parse and handle MIDI messages, so that we reduce note on latency to 0. This would mean wiring up AMY to the MIDI hardware on a device and having APIs that allow for note stealing, etc. 

This could be unnecessary if we find a way to get Python to respond in a predictable amount of time. We can try this out on AMYboard.

dpwe:
Fixed by #323.

PR #323, opened 2025-04-29
AMY MIDI, I2S and instruments / synths

bwhitman:
This is a big change for AMY to handle MIDI, i2s and "synths" natively, instead of relying on other platforms to set that up.

- src/instrument.c: Implements voice-stealing interface.
- instrument.c, patches.c, amy.c: Add  syntax for voice management.
- README describing 'synth=' syntax.
- adding midi in
- note on/off into instrument
- amy tests pass
- stubs for MIDI out, not wired in
- src/midi.c: Add rounding of MIDI out values.
- adding internal field source to prevent MIDI loopback per dan
- synth.source defaults to EVENT_NONE
- hooks to allow external code to access amy midi in and out
- better named hooks
- config
- removing sequencer globals
- moving parse to separate file & putting echo, reverb and chorus in global
- cleaning up delay
- adding separate parse.c
- some cleanup
- using block count instead of sample count
- working tulip midi control of instruments


dpwe (approved):

dpwe:
This addressed much of #302

Issue #321, opened 2025-04-28
Clean up of semantics around "patch"


dpwe:
"Patch" is used several ways within AMY at present:

* A predefined set of parameters for one or multiple oscillators that can be assigned to a voice
* A preset PCM sample
* A preset set of partials "sample"

I think the first is the most important and precise, and the PCM/partials presets should be called something else (maybe `preset`?)

The commands that involve the voice configuration patches are:

* `amy.send(synth=X, voices='a,b,c', load_patch=patch_num)` to set up a block of voices
* `amy.send(store_patch='number,wire_protocol_message')` to define a patch at run time

I propose new options (for the Python interface) to replace existing `patch`, `store_patch`, and `load_patch`:

* `preset` to specify presets for PCM and partials. 
* `patch` to refer to a wire-command string initializing one or more oscillators
* `patch_number` to reference or (re-)define a patch.

So then:

* `amy.send(patch_number=1024, patch='v0S0Zv0S1Zv1w0f0.25P0.5a0.5Zv0w0f261.63,1,0,0,0,1A0,1,500,0,0,0L1')` to define a patch
* `amy.send(synth=0, voices='0,1,2,3', patch_number=1024)` to configure 4 voices under synth 0, or just
* `amy.send(synth=0, voices='0,1,2,3', patch='v0S0Zv0S1Zv1w0f0.25P0.5a0.5Zv0w0f261.63,1,0,0,0,1A0,1,500,0,0,0L1')` to bypass manually assigning a patch number
* `amy.send(osc=0, wave=amy.PCM, preset=11)` to configure a PCM sample
* `amy.send(osc=0, wave=amy.PARTIALS, preset=5)` to configure a preset partials sample

dpwe:
Fixed by #344.

PR #344, opened 2025-05-15
Patch string

dpwe:
This PR includes a bunch of visible changes because several keywords and some message syntax is modified.  Per #321 we make a distinction between `preset`s (the built-in PCM and Partials data, formerly known as `patch`), and `patch` (a wire command string that sets up one or more oscs to serve as a voice).  We also transparently allocate a user patch number (1024-1055) to a `patch` string (formerly the part of a `store_patch` command after the first comma) based on availability (and checking for repeats), so the two stage `store_patch` .. `load_patch` sequence is now a single `amy.send(synth=0, num_voices=4, patch='v0c1A0,1,1000,0,100,0Zv1f510A0,1,500,0,0,100')`.


bwhitman:
<3

PR #355, opened 2025-05-20
renames instrument in the API only to synth

bwhitman:


dpwe:
Looks good, but there's also all of instruments.c which should (at some point) be renamed synth.c, and all its functions should be renamed to s/instrument(s?)_/synth\1_/.

dpwe (approved):
LGTM.  There's still instrument.c, but we can fix that in a separate PR.

PR #363, opened 2025-05-23
API / renaming

bwhitman:
A first pass of refactoring the names we use within AMY and for the API

 - C accessible public functions are now only in `api.c`. 
 - `struct event` is now `amy_event`
 - No more (internal) `amy_add_event`. Now, `amy_add_event(amy_event* e)` is a public C API and the only way to add a C filled event to AMY
 - `amy_parse_event_to_deltas` is now renamed slightly to `amy_event_to_deltas_then`, to indicate the side effect of the callback we set up here
 - What _was_ `amy_add_event` is now just a call to `amy_event_to_deltas_then` (what it actually did)
 - `amy_play_message` is now `amy_add_message`, and just parses the message then adds it
 - There's an internal only `amy_process_event` that has to be called after parsing an event or before adding one. The public APIs do this for you 
 -  Fixed `examples.c` to use `amy_add_event`

So the external API is now just:

* `void amy_add_message(char *message);`: parses and adds a wire string to AMY
* `void amy_add_event(amy_event *e);`: adds an `amy_event` to AMY
* `void amy_parse_message(char * message, amy_event *e);`: just parses a wire string into an event, does not add it
* `void amy_start(amy_config_t);`: starts AMY (not live audio) with given config
* `void amy_stop();`: stops AMY (note: #364)
* `void amy_live_start();`: starts audio playback (i2s, usb, miniaudio, web audio)
* `void amy_live_stop();`: stops audio
* `int16_t * amy_simple_fill_buffer() ;`: manually returns a buffer of audio, if you did not do `amy_live_start()`
* `void amy_update();`: needs to be called in a loop() from MCUs during `live` playback
* `amy_config_t amy_default_config();`: returns default config
* `amy_event amy_default_event();`: returns default empty `amy_event`
* `uint32_t amy_sysclock();`: returns current sample time in ms based on sample clock

And a few things needed by e.g. Tulip Web / AMYboard:

* `void amy_get_input_buffer(output_sample_type * samples);`: access audio input buffer
* `void amy_set_external_input_buffer(output_sample_type * samples);`: set EXT0/EXT1
* `extern uint8_t (*amy_external_render_hook)(uint16_t, SAMPLE*, uint16_t);` for 3rd party code to analyze an oscillator 
* `extern float (*amy_external_coef_hook)(uint16_t);`: for 3rd party code to set two external CTRL COEFs
* `extern void (*amy_external_block_done_hook)(void);`: for 3rd party code to know when a block is rendered
* `extern void (*amy_external_midi_input_hook)(uint8_t *, uint16_t, uint8_t);`: for 3rd party code to receive MIDI data coming into AMY


bwhitman:
( Depending on your style guide, it may be nice to rename all non public APIs like e.g. `amy_process_event` to `_amy_process_event` )

bwhitman:
Fixing the errors, we can tackle the bigger issue on what a "delta" is vs an "event" next

PR #408, opened 2025-06-02
Config is flags; USB support for teensy

bwhitman:


examples/AMY_MIDI_Synth/AMY_MIDI_Synth.ino
```diff
@@ -50,9 +44,6 @@ void setup() {
   amy_config.has_reverb = 1;
   amy_config.has_echo = 1;
 
-  // If you want the default synths (Juno on channel 1, drums on channel 10)
-  amy_config.set_default_synth = 1;
-  amy_config.startup_bleep = 1;
```
dpwe, line 55: I don't know (offhand) what the defaults are (except I think startup_bleep defaults to off).  My point here is that setting them, even if redundant with the defaults, is helpful to the reader/user for (a) being explicit and (b) providing awareness.  To do this with a bit mask is less readable, i.e.
```
amy_config.features |= AMY_FEATURE_DUALCORE;
amy_config.features &= ~AMY_FEATURE_AUDIO_IN;
...
```

examples/AMY_MIDI_Synth/AMY_MIDI_Synth.ino
```diff
@@ -50,9 +44,6 @@ void setup() {
   amy_config.has_reverb = 1;
```
dpwe, line 44: You got rid of these `has_*` flags too, right?  I guess you didn't test this bit on Arduino?

src/AMY-Arduino.h
```diff
@@ -3,9 +3,9 @@
 
 #ifndef AMYARDUINOH
 #define AMYARDUINOH
+//#include <Adafruit_TinyUSB.h>
 
-
-#include "Arduino.h"
+//#include "Arduino.h"
```
dpwe, line 8: This makes me nervous.  We may not need it, but we should be able to include it.

src/amy-piano.c
```diff
@@ -1,8 +1,7 @@
 // amy-example.c
 // a simple C example that plays audio using AMY out your speaker 
+#if !(defined(ESP_PLATFORM) || defined(PICO_ON_DEVICE) || defined(ARDUINO) || defined(__IMXRT1062__) || defined(ARDUINO_ARCH_RP2040) ||defined(ARDUINO_ARCH_RP2350))
 
-// This doesn't compile under ARDUINO because of no usleep.  Arduino tries to compile everything anyway.
-#ifndef ARDUINO
```
dpwe, line 5: I think the only problem is ARDUINO.  It's true we don't want to make this file on the other MCU platforms, but we can achieve that through makefile rules (i.e., not including it).  It's only a problem for Arduino where there is no Makefile and the environment insists on compiling *every* *.c file in the src/ directory.

src/api.c
```diff
@@ -254,10 +251,6 @@ void amy_start(amy_config_t c) {
...
     sequencer_start();
     oscs_init();
-    if(amy_global.config.set_default_synth) {
-	amy_default_setup();
-	if (amy_global.config.startup_bleep) bleep_synth(0);
-    } else {
-	if (amy_global.config.startup_bleep) bleep(0);
-    }
+    if(AMY_HAS_DEFAULT_SYNTHS)amy_default_setup();
+    if(AMY_HAS_STARTUP_BLEEP) bleep_synth(0);
```
dpwe, line 255: No, that's not the logic.  `bleep_synth` can only be used if `HAS_SYNTH`, otherwise we have to use `bleep`.  But if you have `SYNTH` you shouldn't use `bleep` (because it scribbles on an osc without regard to any `SYNTH` setup).  Actually, I changed `bleep` to use `osc = AMY_OSCS - 1`, so maybe it's OK to use in both cases.

src/libminiaudio-audio.c
```diff
@@ -218,13 +218,13 @@ void *miniaudio_run(void *vargp) {
 
 #ifdef __EMSCRIPTEN__
 void amy_live_start_web_audioin() {
-    amy_global.config.has_audio_in = 1;
+    amy_global.config.features &= AMY_HAS_AUDIO_IN;
```
dpwe, line 221: |=

dpwe (approved):
I'm not sold on the features bit mask, but whatevs.

examples/AMY_MIDI_Synth/AMY_MIDI_Synth.ino
bwhitman: fixed with bitfields

src/api.c
bwhitman: Not sure what the fix is here if any?

PR #410, opened 2025-06-04
Examples.py and patch.py

bwhitman:
Ports the C examples to python, and also makes some of the old "presets" into patch.py using AMY's new `patch_number` patch storage.


examples.py
```diff
@@ -0,0 +1,202 @@
...
+
+def example_sine():
+    amy.send(freq=[440], wave=amy.SINE, vel=1)
+
+def example_multimbral_fm():
+    notes = [60, 70, 64, 68, 72, 82]
+    for i, note in enumerate(notes):
+        # Two amy sends, one to load the patch, one to play it
+        amy.send(voices=[i], patch_number=128+i)
+        amy.send(voices=[i], note=note, vel=0.5, pan=[i*2])
```
dpwe, line 131: Those are some big pan numbers.  Do you mean like i % 2?

patch.py
```diff
@@ -0,0 +1,53 @@
...
+        return Patch.patch_num-1
+
+    def synth_with_patch_func(func,synth=0):
+        p = func()
+        amy.send(synth=0, patch_number=p)
+        return synth
+        
+    def simple_sine(**kwargs):
+        p = Patch.new_patch_num()
+        amy.send(osc=0, wave=amy.SINE, bp0="10,1,240,0.7,500,0", patch_number=p,**kwargs)
```
dpwe, line 18: I think it reads better if you put the `patch_number` as the first arg, as it kinda changes the meaning of everything following

dpwe (commented):
So.. the idea of `class Patch` is that it implements the simple examples you had before, but now each one is an instance of `class Patch`, which allocates a new `patch_number` and returns it?  And maybe you can directly set it into a synth using a `function_with_a_long_name`?

In the future, humans do not care about patch numbers.  We almost got rid of them with `patch_string` directly in load-patch statements.  For now, why not have all your examples simply setup the specified synth?  Which defaults to zero, but in the examples have the user specify it?

Hey, these class functions don't have self.  Do they even work?  I don't think they really want to be a class at all, what are the shared features?  You want the thing keeping track of the next unused patch number (or do you? You can keep reusing 1024, the synth you configured doesn't need to look at the patch again), but otherwise they're just functions.  

I guess the thing they have in common is that they can be used as args to synth setup functions.  So like midi.config.add_synth(channel=1, patch=patch.filter_bass()).  Which is easiest with wire strings, but could be something like:
```
def add_synth(self, channel, patch):
  if hasattr(patch, 'patch_number'):
     self.synth_per_channel[channel] = synth.PatchSynth(channel=channel, patch_number=patch.patch_numer)
...
```
or something like that, so that the Patch object has a `patch_number` property which is understood by things that want to consume patches.

bwhitman:
> Hey, these class functions don't have self. Do they even work?

yes

PR #411, opened 2025-06-06
move Juno CC handler to midi.c, check for Juno synth installed first

bwhitman:


bwhitman:
This is all doc changes, except for the Juno CC setup that dan ok'd over iMessage.

Issue #422, opened 2025-06-09
`midi_mappings` is not currently used


dpwe:
`src/midi_mappings.c` was recently introduced as a separate file that could be customized for user-specific MIDI control code interpretation.  It provides code that allows MIDI CCs 70 and 71 to control the VCF base frequency and resonance for Juno voices.

However, its function `juno_filter_midi_handler()` isn't currently used by any AMY build (native or Arduino).  Instead, the Juno filter controls are effected by `amy_received_control_change()` in `src/midi.c`, which duplicates the code from `src/midi_mappings.c` for any channel running a built-in Juno preset patch.

We should remove the example ControlCode handling from `src/midi.c`, and install `juno_filter_midi_handler()` when we call `amy_default_setup()` from `src/api.c`.  My preference would be to move `amy_default_setup()` into `src/api.c`, and include the midi handler installation within it.

dpwe:
Fixed by #426

PR #426, opened 2025-06-10
Enable src/midi_mappings.c, move amy_default_synths() to api.c

dpwe:
Also add Bank Select (CC0) and Volume (CC7).


dpwe:
fixes #422

PR #499, opened 2025-12-13
Fix error when echo delay lines array is empty or partially initialized

TuriSc:
The config_echo function is checking only echo_delay_lines[0] to determine if initialization is needed, but then unconditionally dereferences all channels in the array when setting fixed_delay. This causes en error when the array is partially initialized or contains NULL entries for some channels.


dpwe:
Thanks a lot for raising this; the alloc/dealloc of these large objects was a bit sloppy.  I tightened it up in #500, I hope this has fixed the bug you observed.

Issue #497, opened 2025-12-09
File transfer to host


bwhitman:
We support PCM sample loading via AMY messages. This is very convenient for also sharing files (python) to a host running AMY. We should make a new AMY code to support file loading of tar files that then get unpacked on the host file system . 

for sample loading we use

`z: Signal to start loading sample. preset number, length(samples), samplerate, midinote, loopstart, loopend. All subsequent messages are base64 encoded WAVE-style frames of audio until length is reached. Set preset and length=0 to unload a sample from RAM.`

For file transfer we can use the same code path but

`zT: Signal to start loading file. File size, Destination name, reboot=[0],1 if you want to reboot the host after completing.` 

e.g.

`zT1242,/sd/thing.tar,1` will start a transfer for a tar called thing.tar file of size 1242 bytes, extracted to /sd/thing, and reboot after it's unpacked. 

`z492,/sd/test.py,0` will transfer a 492 byte file test.py to /sd/test.py.

dpwe:
Good point about stereo.  I think we should probably store these samples as stereo by default, perhaps with an option to cast to mono.

Where does the stereo come from?  From the system output.  Rather than sampling `AUDIO_IN[01]`, you sample `BUSx` (in the sense of #114).  For now, that is sampling the system output, which you can set up to simply reflect audio-in, but you can also manipulate along the way.  In the future, you can configure a distinct bus and use it only for sampling.

bwhitman:
OK, so, build this as spec, but also do:

 - new `wave` types of `PCM_LEFT` and `PCM_RIGHT`. We also make a `PCM_MIX` that is aliased to original `PCM` in the enum. 
 - `z` (`pcm_load`) can accept `num_channels` where if 2, the incoming frames are understood to be interleaved L/R and we rename `num_samples` to `num_frames` for clarity. 
 - We update the `memorypcm` structure to handle two channel PCM, with a new `num_channels` field, and interleaving / mixing to a single oscillator based on the oscillator's `wave` type. 
 - `zF` (sets sample from disk) will detect # of channels from wave file header and fill in `memorypcm` channel count, so that a user could use `zF`, then set up stereo playback with `osc=0, wave=amy.PCM_LEFT, patch=1024` and `osc=1, wave=amy.PCM_RIGHT, patch=1024`. Maybe they can set up a two oscillator voice with chained oscillators to treat it as one voice. (and maybe we show examples of this)
[...]

PR #502, opened 2025-12-21
PCM #497

bwhitman:
Does all the stuff in #497.  With tests! And documentation !


bwhitman:
> How about the setting max_value to 1 in PCM. Is that correct and that was just normal snark, or something to fix? The issue was samples with blocks of silence stopping too early.

We talked offline, for now we keep max_value at 1 for PCM. Later, we should have this test in the 0 amp check in amy's renderer (e.g. if wave==PCM && preset==user_preset, then ignore max_value)

PR #509, opened 2026-01-02
We were mallocing full fat instrument struct, not pointers to them, in this setup

bwhitman:


dpwe:
Subsumed by c179774e97783be89f9d116d9762eb84b0caf8cd

Issue #551, opened 2026-02-13
Redefinable sample rate when used as a library


jgartrel:
When using AMY as a library for different boards and use cases it would be nice to be able to modify the sample rate without having to change or fork the AMY library to make that change. 

It looks like all the following defines may be good candidates to be externally configurable:

- AMY_SAMPLE_RATE
- AMY_BLOCK_SIZE
- BLOCK_SIZE_BITS
- PCM_AMY_SAMPLE_RATE

PR #552, opened 2026-02-13
Allow for sample rate to be changed when used as a library

jgartrel:
This pull request allows for the ability to re-define key parameters within AMY without actually modifying the core AMY code. This allows AMY to be adapted for many platforms and use cases when used as a library. Resolves issue shorepine/amy#551

### DIFF
--- a/src/amy.h
+++ b/src/amy.h
@@ -43,6 +43,7 @@ extern const uint16_t pcm_samples;
 
 
 // Set block size and SR. We try for 256/44100, but some platforms don't let us:
+#ifndef AMY_BLOCK_SIZE
 #ifdef AMY_DAISY
 #define AMY_BLOCK_SIZE 128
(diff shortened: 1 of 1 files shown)


jgartrel:
Also, The limited code and changes provided in this pull request are provided freely and completely to the shorepine/amy project to be used as its maintainers see fit and under any copyright they deem approprate.

jgartrel:
@bwhitman , You are right ... I tried to include it this way and it does NOT work, let me do some research and see if there is a way to accomplish this in a sustainable way.

```
#define AMY_SAMPLE_RATE 31250
#include <AMY-Arduino.h>
```

jgartrel:
No doubt you all tested this as a runtime arg.  I just don't see how the compiler could fully optimize in such a case.

Also I want to include a reference to how to extend a core and build properties to include such defines for future people that go down this path (assuming this pull request is accepted): https://github.com/arduino/arduino-cli/issues/1684#issuecomment-1058434618

PR #563, opened 2026-02-24
Add Linux ALSA MIDI support and amy-synth target

bwhitman:
## Summary
- Integrate Alessandro (A.Doinel)'s Linux MIDI contribution using ALSA raw MIDI.
- Add src/linux_midi.c and wire Linux run_midi(), stop_midi(), and midi_out() through it.
- Add src/amy-synth.c and src/alsa_raw_test.c from the submission.
- Update Makefile to build amy-synth; on Linux it now links -lasound and compiles src/linux_midi.c.

## Validation
- make -j4 amy-synth (passes on macOS; Linux ALSA runtime path was not exercised in this environment).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # Makefile for AMY , including an example
 
-TARGET = amy-example amy-message amy-piano
+TARGET = amy-example amy-message amy-piano amy-synth
 LIBS =  -lm  -pthread
 
 UNAME_S := $(shell uname -s)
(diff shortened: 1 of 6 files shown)


bwhitman:
Folded the Linux ALSA MIDI half of this into #1120, which also adds port selection on top of it, so `amy_midi_port_count/name/selected/select` work on Linux as well as macOS. Kept the virtual-port arrangement from here as the default (it is what -1 means on Linux), and fixed three things on the way in: `midi_out` was not calling `midi_out_external_hook`, it wrote to the tx handle with no NULL check, and a port that would not open called `exit(1)`.

Left behind, so this stays open for them: `src/amy-synth.c` no longer links against current amy (`event_generator_for_patch_number` and `print_event` are gone — `sprint_event` is the shape now), and `src/alsa_raw_test.c` is a standalone program nothing builds.

PR #568, opened 2026-03-01
Clear MIDI CC mappings when loading a new patch

bwhitman:
## Summary
- Adds `midi_clear_channel_mappings(channel)` to remove all MIDI CC mappings for a given synth channel
- Calls it from `patches_load_patch()` when a synth is being reset, so stale CC mappings from the previous patch don't persist

## Test plan
- [x] `make test` passes (71/71, one missing ref file for TestCopyingSynthConfig)
- [x] Verify in tulipcc: set MIDI CC mappings on a synth, reload patch → CC mappings should be cleared

### DIFF
--- a/src/patches.c
+++ b/src/patches.c
@@ -1024,6 +1024,8 @@ uint8_t patches_voices_for_load_synth(amy_event *e, uint16_t voices[]) {
             instrument_release(e->synth);
             // Delete the instrument number so we don't forward the 'rest' of the event to it.
             AMY_UNSET(e->synth);
+            // Clear all the midi control code mappings.
+            midi_clear_channel_mappings(e->synth);
             return 0;
         }
     }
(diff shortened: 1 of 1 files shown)


src/patches.c
```diff
@@ -1019,42 +1037,57 @@ uint8_t patches_voices_for_load_synth(amy_event *e, uint16_t voices[], uint16_t
...
+    // This means to set/reset the voices and load the messages (from ROM or memory) and set them.
     peek_stack("load_patch");
     uint16_t voices[MAX_VOICES_PER_INSTRUMENT];
     uint8_t num_voices = 0;
+    uint16_t oscs_per_voice = 0;
     uint16_t patch_number = e->patch_number;
     //fprintf(stderr, "load_patch synth %d patch_number %d num_voices %d\n", e->synth, e->patch_number, e->num_voices);
     if (AMY_IS_SET(e->synth)) {
-        num_voices = patches_voices_for_load_synth(e, voices, &patch_number);
+        midi_clear_channel_mappings(e->synth);
```
dpwe: Actually I think this is misplaced. I need to think more about when exactly we should clear the cc mappings, but for now we reset a synth by explicitly setting it to zero voices.

PR #672, opened 2026-04-07
Add zD, zR, zP wire commands for AMYboard web control mode

bwhitman:
## Summary
- **zD**: Dump state (base64-encoded) or file contents over MIDI sysex. Calls `update_file_hook` before reading to inject fresh AMY state.
- **zR**: Factory reset via platform hook (writes default sketch, restarts)
- **zP**: Ping diagnostic (short sysex reply for connection testing)
- **amy_dump_state_to_string()**: New C API returning all instrument state as a string
- **midi_out**: Loops with `mp_usbd_task()` to drain USB MIDI FIFO for large sysex on ESP32
- **sequencer**: MIDI Start/Stop (0xFA/0xFC) always honored regardless of external clock sync
- **Emscripten fwrite**: Implemented via `amy_shared_write` JS hook (was a no-op)
- **New config hooks**: `amy_external_update_file_hook`, `amy_external_factory_reset_hook`
- **Godot stub** for `midi_out`

## Test plan
- [x] 82 AMY tests pass
- [ ] Hardware: zW writes state file, zD sends it back via sysex
- [ ] Hardware: zR resets to defaults
- [ ] Web: Simulate mode save/load via amy_shared_write hooks
- [ ] Control mode: full sync/save/reset round-trip


src/parse.c
```diff
@@ -440,6 +441,61 @@ uint16_t amy_parse_transfer_layer_message(char *message) {
...
+        // message already points past 'D' (consumed above).
+        // We need to consume all chars up to (but not including) 'Z' so the
+        // outer parser's _next_alpha lands on 'Z' (end of message).
+        // Return value = chars consumed from `message` (after 'D').
+        // The outer parser adds +1 for 'z', so total skip = 1(z) + 1(D via cmd) + len.
+        // BUT the outer parser only does ++pos (for 'z') then _next_alpha which
+        // stops at alpha chars.  To avoid misparse we must consume up to 'Z'.
+        char filename[MAX_FILENAME_LEN];
+        uint16_t len = 0;
+        while (message[len] && message[len] != 'Z' && len < MAX_FILENAME_LEN - 1) {
```
dpwe: (as discussed out-of-band) what to do with filenames containing 'Z'?  I think it's OK to say once you see zD, everything else in the message is taken as a filename and there is no terminating Z (otherwise, what if I want my file called 'Z'?)

dpwe (approved):
I'm nervous about these "big" (16KB) allocs, but if it works, it works.

src/parse.c
bwhitman: We now allow filenames that contain Z. Filenames that end in Z (including a file called Z) will fail. We've added a note in the docs about it. Fixing that is possible but not a "right now on fire" thing

src/transfer.c
```diff
@@ -446,6 +454,137 @@ b64_decode_ex (const char *src, size_t len, b64_buffer_t * decbuf, size_t *decsi
...
+    }
+    if (!amy_global.config.amy_external_fopen_hook ||
+        !amy_global.config.amy_external_fread_hook  ||
+        !amy_global.config.amy_external_fclose_hook) {
+        fprintf(stderr, "zD: file I/O hooks unavailable\n");
+        return;
+    }
+    uint32_t fh = amy_global.config.amy_external_fopen_hook((char *)filename, "r");
+    if (!fh) { fprintf(stderr, "zD: could not open '%s'\n", filename); return; }
+    int max_raw = (MAX_SYSEX_BYTES - 8) * 3 / 4;
```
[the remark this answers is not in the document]
bwhitman: This is now fixed with a max of 96KB per sketch (which can be increased but needs more webmidi fiddlling)

Issue #114, opened 2024-03-30
Buses proposal 


bwhitman:
We have a couple of quality of life issues that a "bus" could help us obviate

 - some patches use chorus, some do not. chorus is currently synth-wide. so it makes it hard to do a multimbral operation
 - some patches use EQ, some do not
 - we would like to route some signals to non-audio ports (DACs for CV operation) 
 
Let's add a bus construct, where oscillators are rendered into a bus and effects can be added after summing into that bus. 

If this sounds complicated, never fear, we already have two buses! `fbl` is used on multi core hardware to render a group of oscillators. We then sum those and add effects to them together. What if `fbl` could be, say, up to 6, instead of two, and you can assign oscillators to a bus number. By default, they would be set up like they are now, into two buses by their osc #, `(0-AMY_OSCS/2, AMY_OSCS/2-AMY_OSCS)`. But a parameter, say, `U` / `bus`, can be added to the oscillator to override this bus.

Then, synth wide "effects" like (currently) chorus, reverb, eq can also have `U` / `bus` set alongside them (with default being all active buses), e.g. `chorus_level=1, bus=0,1,3`. And (for example) juno patches can have bus numbers pre-set into ones that need chorus / eq / etc and those that don't. Then dx7 and juno voices could be played simultaneously with chorus only on the appropriate ones.

(Future) a bus can be indicated as non-audible, or have its volume set per bus for mixing. Other C code could peek into a bus to, for example, send the bus output to a non-audio sink, like a i2c DAC on Tulip for modular synth ramps.

bwhitman:
A more concrete buses spec:

We make a new parameter `Y = bus_input`. `bus_input` can be 0-7 (internal rendering buses), or 8-15 (external input buses). `bus_input` is 0 by default (if not given)

We also make a new parameter `z = bus_output`. This is 0 by default (first stereo output) but can be up to 7.  

Ignoring audio in for now, imagine a Juno-6 patch that has `Y1` in it. This would mean that any (eg) chorus, EQ, or reverb command only turns on for that bus. We also make a `Y2` for every dx7 patch, so that those effects do not render for those patches. We implement this by having more fbl in our rendering (instead of 2, we can have 8 of them) 

Now, I want the audio in to go through an echo. I send AMY `Y8M1`. This tells AMY that bus 8 (first external input) goes through echo with level 1.0. I can render effects on any input. 

[...]

PR #686, opened 2026-04-18
Buses

dpwe:
Implements most of #114 

This adds a new "bus" field to `amy_event` and `struct synthinfo`.  `amy_global` now has an array of `bus_state_t *bus[AMY_NUM_BUSES]` each of which holds params (and pointers to private buffers) for EQ, reverb, chorus, and echo.

`amy_render` looks at the `bus` field of the "lead osc" for each note, and calls `render_osc_wave` with a pointer to the bus-specific buffer, thus collecting the output for each bus separately (we now have `AMY_NUM_BUSES * AMY_MAX_CORES` buffers).

`amy_fill_buffer` executes EQ, Chorus, Echo and Reverb separately for each active bus, then sums the results across buses, scaled by `amy_global.volume[bus]`, before converting to the final output samples.  So, `volume='<bus0_level>,<bus1_level>,...'` is now a list.

Apart from converting all the global buffers for the FX into resources managed by a struct, the weirdest thing I had to do was to hijack the `d->osc` field in `struct delta` to hold the value of the *bus* (instead of the osc) for commands that are setting the params of a bus - otherwise, there's no way to combine both the parameter value with the bus to which it is meant to apply in a delta.  So there's a big explicit conditional in `amy_event_to_deltas_queue` that treats all the EQ, echo, chorus, and reverb commands as a special-case.  It's a bit clumsy, but not a bad solution.  Maybe we should rename `d->osc` to be `d->target` since it's not always an osc now, but that's just to avoid misdirecting readers of the code.

This went so much more smoothly than I expected.  One 4 hour session to move all the FX state into structures so they could be multiply-allocated, then another 4 hour session to actually implement the per-bus processing.


dpwe:
# TESTED:
I reflashed AMYboard (with [current tulipcc head](https://github.com/shorepine/tulipcc/commit/411daee036f7f29dfaf5b53231a02ae94932ce0c)) using this patch and everything seems OK.  I tried directly manipulating buses to check it was working; after setting synth 2 to be a DX7 voice, I used the REPL (connecting via `screen`) to move it between bus 0 and bus 1, with reverb turned up for bus 0 but not bus 1, and verified that switching between buses caused the synth to render with and without reverb.  I also tested that the multi-parameter `volume` command appropriately affected different buses differently:
```
>>> amy.send(bus=0, reverb=1)  # play notes (on MIDI channel 2), synth is on bus 0, hear reverb
>>> amy.send(synth=2, bus=1)  # play notes, synth is on bus 1 which has no reverb, hear no reverb
>>> amy.send(volume='5,1')  # Adjust volume of buses 0 and 1; synth is quieter
>>> amy.send(volume='1,5')  # Flip gains; synth is louder
>>> amy.send(volume='5,5')  # Make gains same
>>> amy.send(synth=2, bus=0)  # Move synth back to bus 0; still louder, now has reverb too.
```

Issue #625, opened 2026-03-22
ESP-IDF: AMY crashes at boot — ESP_TASK_PRIO_MAX is invalid as a FreeRTOS task priority


rt-rtos:
# Bug: `AMY_RENDER_TASK_PRIORITY` crashes on ESP-IDF 6.0 — `ESP_TASK_PRIO_MAX` equals `configMAX_PRIORITIES`, which is an invalid FreeRTOS priority

## Environment

| Field | Value |
|---|---|
| Framework | ESP-IDF 6.0 (not Arduino) |
| Target | ESP32-S3 |
| AMY version | latest `main` (as of 2026-03-22) |

## Problem

AMY crashes immediately at boot with:

```
assert failed: prvInitialiseNewTask tasks.c:1111 (uxPriority < ( 25 ))

Backtrace:
--- 0x4200fe9e: amy_platform_init at components/amy/src/i2s.c:293
--- 0x4200fc31: amy_start at components/amy/src/api.c:361
```

The crash happens inside `amy_platform_init` → `xTaskCreatePinnedToCore` in `src/i2s.c:293`, before any audio initialisation occurs. The device boot-loops unconditionally.

## Root Cause

In `src/amy.h`, the non-Arduino ESP-IDF path defines:

```c
// src/amy.h (non-Arduino path)
#define AMY_RENDER_TASK_PRIORITY      (ESP_TASK_PRIO_MAX)
#define AMY_FILL_BUFFER_TASK_PRIORITY (ESP_TASK_PRIO_MAX)
```

`ESP_TASK_PRIO_MAX` is defined by ESP-IDF as:

```c
// esp_system/include/esp_task.h
#define ESP_TASK_PRIO_MAX (configMAX_PRIORITIES)   // = 25
```

FreeRTOS's `prvInitialiseNewTask` asserts **strict less-than**:

```c
// FreeRTOS-Kernel/tasks.c:1111
configASSERT( uxPriority < configMAX_PRIORITIES );
```

Passing `configMAX_PRIORITIES` (25) is therefore **always invalid** on any standard ESP-IDF build and crashes unconditionally. This is not a configuration issue — it is a hardcoded invalid value.

### The SDK documents the correct pattern

The `esp_task.h` header itself carries this comment immediately above the definition:

> *otherwise use `ESP_TASK_PRIO_MAX - X` style*

Every other ESP-IDF internal task follows this convention, e.g.:

```c
[...]

PR #692, opened 2026-05-05
src/amy.h: Make RTOS priorities strictly < ESP_TASK_PRIO_MAX for #625.

dpwe:
Proposed fix for #625 .  Reduces RTOS priorities for AMY `render` and `fill_audio_buffer` tasks to 24 (`ESP_TASK_PRIO_MAX - 1`) because 25 causes an assert failure on ESP-IDF 6.0.

Untested as yet (I do not have ESP hardware to hand).


dpwe:
# AMYBOARD
Flashed onto AMYboard using TulipCC [`411daee`](https://github.com/shorepine/tulipcc/commit/411daee036f7f29dfaf5b53231a02ae94932ce0c), all seems to work fine.  6 simultaneous patch 1 voices with reverb, no dropouts.

dpwe:
presumed to fix #625

PR #694, opened 2026-05-05
Make midi_mappings also cover MIDI note-on events.

dpwe:
This takes the midi mapping structure built to automatically convert MIDI Control Change messages into wire commands, and extends it to work for MIDI note-on message.  By default, MIDI note-on messages have the mapping `-1,0,0,1,0,i%in%nl%v`, meaning that any note is converted to a note-on command for the matching synth.  But this command can be overridden e.g. `amy.send(synth=1, midi_note_cmd='-1,0,0,1,0,i%in%nl%vZ-1,0,0,1,0,i2n%nl%v')` would convert any MIDI note-on command on channel 1 to two internal note-ons, one to synth 1, and a simultaneous one to synth 2.


dpwe:
WARNING: This modification removes the setting of `e.note_source = NOTE_SOURCE_MIDI;`.  Whatever relied on this (I think, echoing midi in to midi out) will now be broken.

EDIT: This is fixed in added commit 8597808c21fa8941309d68215fffacebc35d33aa, but I also identified a bug with `NOTE_SOURCE_MIDI` (#695) that once a MIDI note has been seen, the entire synth is marked as MIDI-sourced from then on.

dpwe:
# AMYboard Testing
This PR changes the name and signature of `midi_mappings.c:midi_msg_handler()` which is called from `amy_connector.c`, so I had to make a new branch, [`tulipcc:midi_note_on_mapping`](https://github.com/shorepine/tulipcc/commit/ec302178031efe951d227635cd570364b842cab0) to make it compile.  Using that branch, I reflashed AMYboard and verified that MIDI control still works, both for note on/off and for CC changes.

PR #704, opened 2026-05-30
fix default config in the python module

finalpatch:
Problems fixed:
1. Using uninitialized `amy_global.config`  to call `amy_start()` (`amy_global.config` is initialized inside `amy_start()`)
2. Hard-coded audio input = 1
3. "<strings.h>" header is not always implied on every platform


bwhitman:
thank you!

PR #712, opened 2026-06-04
Fix AMY_arduino for RPi Pico: No sysex alloc, add bus args to config_…

dpwe:
…echo etc.

Confirmed that BillieJeanScheduled and AMY_MIDI_Synth sketches work on Pimoroni Pico2 using RPi Pico Core 5.5.1.


dpwe:
Also confirmed under Arduino 2.3.9, pico-core 5.6.0

PR #725, opened 2026-06-14
Install amyboard default MIDI CC map when patch 257 loads

bwhitman:
## Summary

Patch **257** (the "amyboard default") already sets up the oscillators, but the default MIDI CC assignments only lived in the AMYboard web editor (`amy_parameters.js`, amy #920). So a freshly-booted / factory-reset AMYboard had **no CC mappings** until the website pushed them — a plugged-in MIDI controller did nothing out of the box.

Now **loading `K257` on a channel also installs the standard default CC map on that channel**, so a controller drives the usual parameters (level, VCF cutoff/resonance, ADSR, LFO rate/depth, reverb/chorus/echo) with no `ic` setup.

> Re-do of #723 (which was merged early and reverted in #724). Same change, cherry-picked onto current `main`.

## How

- `midi_install_amyboard_default_ccs(int channel)` (`src/midi_mappings.c`): a table of the 12 default CCs mirroring tulipcc's `tulip/amyboardweb/static/amy_parameters.js`, installed via `midi_store_mapping()`. Templates use the same `%i`/`%v` placeholders as a hand-written `ic` code.
- `patches_load_patch()` (`src/patches.c`): calls it when `patch_number == 257` and a synth is set.

Purely additive, only fires on patch 257. The state-dump path (`amy_emit_state_lines`) emits explicit `iv/in` + osc state + `ic` lines and never `K257`, so saved-sketch replay is unaffected; the web editor already re-sends its CCs after `K257` (idempotent).

## Default CC map

| CC | Param | log | min | max | offset |
|---|---|---|---|---|---|
| 7 | channel level | ✓ | 0.001 | 7 | 0.1 |
| 74 | VCF freq | ✓ | 20 | 8000 | 0 |
| 71 | VCF resonance | ✓ | 0.5 | 16 | 0 |
| 76 | LFO freq | ✓ | 0.1 | 20 | 0 |
| 77 | LFO→osc depth | ✓ | 0 | 4 | 0.001 |
| 93 | chorus level | ✓ | 0 | 1 | 0.1 |
| 73 | ADSR attack | | 0 | 1000 | 0 |
| 75 | ADSR decay | ✓ | 0 | 2000 | 50 |
| 79 | ADSR sustain | | 0 | 1 | 0 |
| 72 | ADSR release | ✓ | 0 | 8000 | 50 |
| 91 | reverb level | ✓ | 0 | 1 | 0.1 |
| 94 | echo level | | 0 | 2 | 0 |

## Test

`TestPreset257DefaultCCs` loads `K257`, injects CC7=30, and confirms the channel level drops with no explicit `ic` setup. Full suite: **93 tests pass** locally.


amy/test.py
```diff
@@ -1238,6 +1238,20 @@ def run(self):
...
+class TestPreset257DefaultCCs(AmyTest):
+  """Loading K257 (the amyboard default) installs a default MIDI CC map, so a
+  controller works with no explicit `ic` setup. Here CC7 (channel volume) is
+  injected before the note to drop the channel level — audible proof that the
+  default mapping is live."""
+
+  def run(self):
+    amy.send(time=0, synth=1, patch=257, num_voices=4)
+    # inject_midi args are (time, midi_status, data1, data2). 0xB0 = CC on ch 1.
+    amy.inject_midi(50, 0xB0, 7, 30)  # CC7 -> i1v0a..., lowers the channel level
```
dpwe, line 1250: Did you try this test before the change?  CC 7 for volume may already have been hard-coded at `amy_midi.c:82`.

src/midi_mappings.c
```diff
@@ -165,6 +165,52 @@ int midi_store_mapping(int channel, int type, int code, int is_log, float min_va
...
+// Amyboard default MIDI CC map.
+//
+// Installed for a channel when patch 257 (the "amyboard default") is loaded, so
+// a MIDI controller can drive the standard synth parameters out of the box —
+// before the web editor pushes any `ic` mappings (e.g. on a freshly-booted /
+// factory-reset board). These mirror the default `cc:` assignments in tulipcc's
+// tulip/amyboardweb/static/amy_parameters.js (introduced in amy #920); keep the
+// two tables in sync. Templates use %i (channel) and %v (value) exactly like a
+// hand-written ic code, and v0/v1/v2/v3 are the K257 control/LFO/oscA/oscB oscs.
+struct amyboard_default_cc {
```
dpwe, line 177: I would probably have done this as straight wire strings, e.g. `sprintf(str, "i%dic7,1,0.001,7,.1,i%%iv0a%%v", channel)`, then `amy_add_message(str)`. Not as readable, but avoids introducing a novel datatype, and keeps `midi_store_mapping` as called only from `parse.c`.  Not sure it's worth changing now.

However, if we just append wire commands like this to the K257 patch in patches.h (i.e., put these wire commands into the output of `headers.py:make_amyboard_patch()`, then we can do this with no changes/special cases in the AMY engine?  That seems worthwhile.

bwhitman:
This was fixed eslwhere

PR #732, opened 2026-06-17
Fix TRUE_EXPONENTIAL envelope step from SAMPLE overflow (#731)

### LINKED ISSUE #731: TRUE_EXPONENTIAL envelope has problems.
The following code generates a tone with a step increase in amplitude after the attack segment completes:

```
amy.reset()
amy.send(osc=0, eg0='0,0.01,8000,1,8000,0,0,0', eg0_type=amy.ENVELOPE_TRUE_EXPONENTIAL)
amy.send(vel=1)
```

With other envelope types (e.g. `ENVELOPE_LINEAR`), it gives a smooth fade-in then fade-out with no jump.

bwhitman:
Fixes #731.

## Symptom

@dpwe reported that `ENVELOPE_TRUE_EXPONENTIAL` produces a step jump in amplitude right after the attack segment completes, where `ENVELOPE_LINEAR` is smooth:

```python
amy.send(osc=0, eg0='0,0.01,8000,1,8000,0,0,0', eg0_type=amy.ENVELOPE_TRUE_EXPONENTIAL)
amy.send(vel=1)
```

TRUE_EXPONENTIAL attack, before vs after

## Root cause

`SAMPLE` is Q8.23 fixed point (`S_FRAC_BITS = 23`), so its maximum magnitude is **256**. The branch built the curve from `rate * (elapsed - t0)`:

```c
SAMPLE dx7_exponential_rate = F2S(... / (0.001f * (t1 - t0)));
scale = MUL4_SS(v0, exp2_lut(MUL4_SS(dx7_exponential_rate,
                                     F2S(0.001f * (elapsed - t0)))));  // overflows
```

The intermediate `F2S(0.001f * (elapsed - t0))` hits `0.001 * 256000 = 256` at **~5.8 s** into the segment (≈5.3 s @ 48 kHz) and overflows Q8.23. Past that point the exponent stops growing, so a *rising* envelope freezes partway (~27% in the report's case) and then snaps to the target at the segment boundary. `LINEAR` is unaffected because it interpolates with the bounded `time_ratio ∈ [0,1]`; exponential *decays* survive because they reach their floor well before the overflow.

## Fix

Interpolate in the log2 domain using the already-bounded `time_ratio` — mathematically identical to the intended exponential, but no term ever exceeds ~13:

```c
float log2_v0 = S2F(log2_lut(v0));
float log2_v1 = S2F(log2_lut(v1));
scale = exp2_lut(F2S(log2_v0 + (log2_v1 - log2_v0) * time_ratio));
```

## DX7 decay — decision for @dpwe (commit 2, droppable)

The DX7 *"decay is regular true_exponential"* branch uses the identical `rate * (elapsed - t0)` formulation and has the **same overflow** for DX7 decays longer than ~5.8 s. The second commit applies the same fix there.

Because that recomputes the (short-segment) DX7 decay slightly differently in fixed point, it shifts **7 calibrated FM/DX7 golden references at the rounding level** (~−75..−97 dB error — inaudible), regenerated in that commit: `TestXanaduFM`, `TestSynthProgChange`, `TestWoodPiano`, `TestFMRepeat`, `TestAlgo2`, `TestAlgo`, `TestSineEnv2`.

It is intentionally **isolated as a separate commit** so it can be dropped if you'd rather not move the FM references. The `TRUE_EXPONENTIAL` fix (commit 1) stands alone with **zero reference changes**.

## Verification

- #731 case: the boundary step drops from **+71% → +0.5%** of full scale; the attack rises smoothly and peaks exactly at 8 s.
- Robust across segment lengths, including a 20 s segment (3.4× the old overflow point).
- `python -m amy.test`: **93/93 pass**.


dpwe:
I rewrote this to avoid the duplicated code in #734

PR #733, opened 2026-06-17
Fix premature termination of oscillators (#720)

### LINKED ISSUE #720: Premature termination of oscs
Deciding when an osc is "done" is a little tricky: You can't just stop the oscs when the note-off occurs, because if the envelope has a nonzero decay, the note will continue for some time after the note-off.  But because of the way `amp` is controlled by a set of CtrlCoefs, it's not straightforward to decide when a note is done - for instance, it could be either, or both, envelope generators that are controlling the release.

The current approach is that the engine monitors the peak amplitude in each generated block, and if the output has been zero for a certain length of time, we decide the osc is done and stop evaluating it until the next note-on.  However, this can mistakenly terminate oscs that would, otherwise, make sound at some point in the future.

Osc startup is a specific case.  We don't wait for a note-off to apply the zero-output condition, it's applied all the time.  We don't wait for an osc to start sounding before we inspect it for zero output.  So an osc that has zero (or very low) output at its onset - for instance, a note with a very slow attack - *can* be terminated before it ever gets going.

Being somewhat aggressive about turning off unused oscs at onset is helpful for Juno voices, which have 4 sounding oscs (pulse, saw, suboctave, and noise) but for many patches only 1 or 2 oscs are actually used (i.e., have nonzero amps).  By turning these off very soon after onset, we reduce the computational load of running Juno voices, and increase overall Juno polyphony for many patches.

For this reason, although the termination condition is 10 frames of zero amp (54 ms), at onset the osc is allowed only a couple of frames (11 ms) to emit some signal, otherwise it is abandoned, presumed inactive.

This policy has more than once caused confusion, since the symptom is notes that simply do not sound.

It's not clear to me how to fix this.  Some options:

* More sophisticated decisions about whether an osc is done or not - possibly looking at the amp CtrlCoefs and deciding which envelopes will control the decay.  (Note how hard this is since the amp can also depend on modulation oscs, or even external CV inputs).
* Longer timeouts before terminating voices.  They're short - particularly at onset - to maximize the polyphony, particularly of Juno patches.  We could actually structure these patches so that each patch only chains the oscs that are actually in-used (this, however, would be defeated by #716).
* More sophisticated treatment specifically of note-ons, so that truly inactive voices waste minimal resources, but conditions that could result in delayed appearance of the amplitude will get more patience.

bwhitman:
Fixes #720.

## Problem

Deciding when an osc is "done" was too aggressive at note onset. The silence detector gave each osc only ~2 blocks (~11 ms) to emit signal before abandoning it and suspending it until the next note-on. An osc that is legitimately silent at onset — a **delayed or very slow attack**, or a momentary dip below the noise floor — could be killed before it ever sounded, producing the "notes that simply do not sound" symptom described in the issue.

## Approach

Implements the plan from @dpwe's comment on #720:

1. **Suspend truly-dead oscs at onset instead of relying on the silence-cull.** At note-on, any osc whose amplitude is a hard constant zero (`amp_coefs[COEF_CONST] == 0`) and isn't amp-modulated (`amp_coefs[COEF_MOD] == 0`) is moved straight to `INAUDIBLE` — it can never make sound, since every other amp control can only attenuate further. This is what keeps lightly-used Juno voices (4 oscs, often only 1–2 sounding) cheap, so we no longer need the aggressive onset cull to claw back the polyphony. The osc is still fully set up (`osc_note_on`), so a later amp change revives it via the existing path.

2. **Don't terminate while the gate is active.** Envelope-gated oscs are protected from silence-termination as long as the note is held (note-on seen, `note_off_clock` not yet set). The terminate-on-silence logic now only runs in the release phase, with the existing ~54 ms timeout.

3. **Walk past inactive oscs in a chain.** Because (1) can now suspend an osc in the *middle* of a chain, `render_osc_wave` skips inactive oscs but keeps walking the chain, so a suspended osc no longer strands the still-sounding oscs behind it (e.g. a Juno VCA → silent saw → sounding sub-osc).

### Design note for review

`note_off_clock` is deliberately not set on note-off for KS / ALGO / PCM / CUSTOM / partials (it would make their `bp0` envelope hard-zero them — see the comment at the old note-off switch). So those waves can't use it as a "released" signal. To avoid leaking silent oscs for them, the gate in (2) is applied only to waves that *do* use `note_off_clock` as their release signal (`wave_terminates_only_in_release()`); everything else keeps the exact legacy terminate-whenever-silent behaviour. Happy to adjust where that line is drawn.

## Testing

- All **93 existing tests remain bit-exact** (`err = -100.0 dB`), including every Juno patch, chained-osc, voice-stealing and sustain test — so the optimization in (1) and the chain change in (3) are output-preserving on real patches.
- **`TestDelayedAttack`** (new): a note whose envelope holds at zero for 150 ms before ramping up. Renders **completely silent on `main`** (`signal = -100 dB`) and sounds correctly with this change — a direct regression guard for the bug.
- **`TestChainSkipsSilencedOsc`** (new): a SILENT VCA → zero-amp pulse (suspended) → sounding saw, locking in that the suspended mid-chain osc doesn't strand the saw.

```
95 tests pass
```


bwhitman:
dan is gonna do this himself

PR #739, opened 2026-06-20
Add amy_config.midi_thru: echo MIDI in to MIDI out (#738)

### LINKED ISSUE #738: MIDI thru
We could have an amy side confiugration to pass all MIDI in to MIDI out, byte by byte, before processing. Perhaps setting at runtime somehow via `amy_config`?

bwhitman:
Closes #738.

## What

Adds an `amy_config_t.midi_thru` flag. When set, AMY echoes **every incoming MIDI byte straight to MIDI out**, in addition to processing it normally. **Off by default.**

```c
c.midi_thru = 1;          // enable at startup
amy_start(c);
// ...or toggle at runtime:
amy_global.config.midi_thru = 1;
```

## How

Forwarding is done byte-by-byte at the top of `convert_midi_bytes_to_messages()` — the single function that all MIDI inputs (UART via `esp_poll_midi`/`on_pico_uart_rx`, USB via `check_tusb_midi`) funnel through.

Forwarding inside the parse loop (rather than blasting the whole input buffer) keeps USB correct: USB MIDI always delivers fixed 3-byte groups, padded with junk for <3-byte messages. The existing `if(usb) i = len+1` early-exits stop the loop before those padding bytes, so only the real MIDI bytes of a USB message are passed through. UART input has no padding and forwards in full.

## Notes

- No new symbols added to `amy_midi.c` (only the body of an existing function changed), so no Godot platform-stub updates are needed.
- On Tulip this is exposed to Python as `tulip.amy_midi_thru(True/False)` (companion tulipcc PR).
- Web targets are out of scope for now; the field is platform-agnostic and stays off unless explicitly enabled.

## Test

- `make test` — all 94 tests pass (midi_thru is off by default; no behavior change to existing paths).
- Verified `amyboard` (ESP32-S3) firmware compiles with the change.

Docs updated: `docs/api.md` (config table) and `docs/midi.md` (new "MIDI Thru" section).

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -155,6 +155,7 @@ amy_start(amy_config);
 | `i2s_lrc`, `i2s_dout`, `i2s_din`, `i2s_bclk`, `i2s_mclk` | Int | -1 | Pin numbers for the I2S interface |
 | `midi_out`, `midi_in` | Int | -1 | Pin number for the MIDI UART pins |
 | `midi_uart` | 0,1,[2] | -1 | UART device index for MCU. Default 1 (`UART1`) on Pi Pico and ESP. Teensy is always `8` |
+| `midi_thru` | `0=off, 1=on` | Off | If set, every incoming MIDI byte is echoed straight to MIDI out, in addition to being processed normally. Can be toggled at runtime via `amy_global.config.midi_thru` |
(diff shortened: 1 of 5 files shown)


bwhitman:
We decided to not support "soft" MIDI Thru at this time, it causes too many issues that a real MIDI thru box will solve for anyone.

PR #741, opened 2026-06-20
Run AMYboard hardware CI on AMY PRs

bwhitman:
## What

Adds a thin trigger so an AMY PR gets tested on the **physical AMYboard bench**, the same way tulipcc PRs already are — without copying any of the bench machinery here.

On a same-repo PR touching `src/**`, this fires a `repository_dispatch(amy-pr)` at **shorepine/tulipcc**, which builds the AMYboard firmware with *this PR's* `amy` submodule SHA, runs the board, and comments pass/fail back on this PR. All the heavy lifting (firmware build, Vercel flasher, self-hosted Pi runner, `hwci.py`, reference audio) stays in tulipcc.

Paired with **[shorepine/tulipcc#1020](https://github.com/shorepine/tulipcc/pull/1020)**, which adds the receiving end.

## Flow

```
this PR (src/**, same-repo) → repository_dispatch(amy-pr) → tulipcc bench → result comment back here
```

## Security

- **Same-repo only** — the job is gated to PRs whose head branch is on shorepine/amy. Fork PRs are skipped so fork code never reaches the self-hosted Pi (a maintainer can still run a fork manually via tulipcc's "AMYboard PR preview" `workflow_dispatch`).
- Fork PRs also can't read the dispatch secret, so they can't fire it even if the gate changed.

## Required before it works

- Repo secret **`HWCI_BRIDGE_TOKEN`** — a token with `contents: write` on shorepine/tulipcc (to send the dispatch). ✅ already added.
- **Merge order:** merge [tulipcc#1020](https://github.com/shorepine/tulipcc/pull/1020) first (the dispatch needs a listener on tulipcc `main`), then this. A `pull_request` trigger for a brand-new workflow only fires on PRs opened *after* it lands on `main`, so the real end-to-end test is the *next* `src/**` PR.

### DIFF
--- a/.github/workflows/amyboard-hwci-trigger.yml
+++ b/.github/workflows/amyboard-hwci-trigger.yml
@@ -1,0 +1,79 @@
+name: AMYboard HW CI (dispatch)
+
+# Run THIS PR's AMY code on the physical AMYboard bench — without duplicating any
+# of the bench machinery. The firmware build, the Vercel preview/flasher, the
+# self-hosted Pi bench, the hwci.py driver, and the reference audio all live in
+# shorepine/tulipcc (AMY is a submodule there). So instead of copying that here,
+# we fire a `repository_dispatch` at tulipcc telling it: build your AMYboard
+# pipeline but pin the `amy` submodule to THIS PR's SHA, run the bench, and
+# comment the pass/fail back on this PR.
+#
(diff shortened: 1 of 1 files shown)


bwhitman:
@dpwe this is not a code change, just a workflow change for CI - i'm going to merge it now

PR #743, opened 2026-06-20
Fix NULL deref on freed chained_osc in render_osc_wave

rt-rtos:
`render_osc_wave` dereferences ` synth[chained_osc]->status` with no NULL check on
`synth[chained_osc]` itself. A chained_osc slot can be freed or left unallocated
during a patch toggle, causing a NULL deref fault 
The fix adds a NULL check before the status check, matching the pattern used
elsewhere for `synth[]` slot access. When` chained_osc` is NULL the `chained osc `is
skipped — it cannot be `SYNTH_AUDIBLE`, so behaviour is unchanged when allocation
is intact.

### DIFF
--- a/src/amy.c
+++ b/src/amy.c
@@ -1702,7 +1702,7 @@ SAMPLE render_osc_wave(uint16_t osc, uint8_t core, SAMPLE* buf) {
         if(AMY_IS_SET(synth[osc]->chained_osc)) {
             // Stack oscillators - render next osc into same buffer.
             uint16_t chained_osc = synth[osc]->chained_osc;
-            if (synth[chained_osc]->status == SYNTH_AUDIBLE) {  // We have to recheck this since we're bypassing the skip in amy_render.
+            if (synth[chained_osc] != NULL && synth[chained_osc]->status == SYNTH_AUDIBLE) {  // We have to recheck this since we're bypassing the skip in amy_render.
                 SAMPLE new_max_val = render_osc_wave(chained_osc, core, buf);
                 if (new_max_val > max_val)  max_val = new_max_val;
             }


bwhitman:
### 🎛️ AMYboard HW CI

bwhitman:
### 🎛️ HW CI (physical bench)

**[⬇️ Artifacts: recordings · screenshot · serial logs · run logs](https://github.com/shorepine/tulipcc/actions/runs/27890079007)**

dpwe:
I can't think of any circumstance when a chained osc would be unallocated.  I'm not clear what a "patch toggle" is.

rt-rtos:
Sorry for the lack of clarity - I'm running render and patch changes on separate RTOS tasks pinned to different cores, so here's the path that produces it:

Edit: This is running multicore but with Amy `multicore=0`. I appreciate this may be outside the intended use case. happy to close if so.

free_osc(X) sets synth[X] = NULL. But it doesn't scan all other oscillators for chained_osc == X and clear them. So if synth[A]->chained_osc points at X, and free_osc(X) runs concurrently with (or just before) the render path reading that reference, AMY_IS_SET(synth[A]->chained_osc) is still true — the index is intact — but synth[X] is now NULL.

I hit this during patch toggling under sequencer load: the crash was EXCVADDR=0x00000008, which is NULL + offsetof(synth_t, status) — the render path got into render_osc_wave, passed the AMY_IS_SET check on the index, then dereferenced synth[X]->status against a NULL pointer.

dpwe:
Thanks for the added detail, I see it now.  It makes me very nervous to think of rendering happening while osc teardown is progressing - I expect there are lots of situations where I assume things exist, and even this check could potentially fail if the 2nd core changes the pointer after the NULL check occurs.  A truly robust solution would involve some locking, or a carefully-designed synchronization so that tear-down can occur only when rendering is idle.  But I hope this change will address the most common problems.

PR #744, opened 2026-06-21
fix: make reverb allocation failure safe (mirrors existing echo guard)

rt-rtos:
### Bug:
`init_stereo_reverb()` allocates 10 delay lines but returns `void`, so a partial
allocation failure is silently ignored. The idempotency guard at the top of the
function checks only rev->delay_1: if delay_1 succeeds but any later line fails,
`rev->delay_1` is non-NULL on all subsequent calls, so the half-initialised struct
is never repaired. `config_reverb()` then sets reverb.level unconditionally, and
the render guard in `amy_fill_buffer` checks only `reverb.level > 0` — so
`stereo_reverb()` runs and dereferences a `NULL` delay-line pointer, guaranteed crash.

Echo already has the correct guard `(echo_delay_lines[0] != NULL)`. This PR brings
reverb in line, adapted to the struct-based `reverb_params_t design`:

**Allocator — returns bool, rolls back on failure:**  
```c
amy.c - line 204

bool alloc_echo_delay_lines(uint8_t bus, uint32_t max_delay_samples) {
    bool success = true;
    for (int c = 0; c < AMY_NCHANS; ++c) {
        delay_line_t *delay_line = new_delay_line(max_delay_samples, 0, amy_global.config.ram_caps_delay);
        if (delay_line) {
            amy_global.bus[bus]->echo.echo_delay_lines[c] = delay_line;
        } else {
            success = false;
            break;
        }
    }
    if (!success) {
        fprintf(stderr, "unable to alloc echo of %d samples\n", (int)max_delay_samples);
        dealloc_echo_delay_lines(bus);
        return false;
    }
    return true;
}
``` 

**config_echo — bails on allocation failure before committing level:** 
```c
amy.c - line 243-245

    if (level > 0) {
        if (amy_global.bus[bus]->echo.echo_delay_lines[0] == NULL) {
            if (!alloc_echo_delay_lines(bus, max_delay_samples)) return;
        }
        ...
    }
    amy_global.bus[bus]->echo.level = F2S(level);
```

**Render guard:** 
```c
amy.c - line 1975
if (amy_global.bus[bus]->echo.level > 0 && amy_global.bus[bus]->echo.echo_delay_lines[0] != NULL) {
````

### Fix:
- `init_stereo_reverb(reverb_params_t *rev)` returns `bool`. On any allocation
  failure it calls the existing `deinit_stereo_reverb()` to roll back all partial
  allocations (leaving every pointer `NULL`) and returns `false`.
- `alloc_reverb_delay_lines()` returns `bool`, propagating the failure to its caller.
- `config_reverb()` forces level=0 and returns early on allocation failure instead
  of committing a non-zero level against an uninitialised rev.
- The render guard adds `rev != NULL && rev->delay_1 != NULL` as a final defence
  against NULL deref (mirrors the echo pattern; delay_1 is the first line
  allocated and the last freed, making it a reliable sentinel).

Behaviour is unchanged when allocation succeeds. The added render-path check is
two pointer comparisons per block.

Pardon me if I am missing something obvious and wasted anyones time.

### DIFF
--- a/src/amy.c
+++ b/src/amy.c
@@ -330,10 +330,10 @@ void config_chorus(uint8_t bus, float level, uint16_t max_delay, float lfo_freq,
     amy_global.bus[bus]->chorus.depth = depth;
 }
 
-void alloc_reverb_delay_lines(uint8_t bus) {
+bool alloc_reverb_delay_lines(uint8_t bus) {
     if (amy_global.bus[bus]->reverb.rev == NULL)
         amy_global.bus[bus]->reverb.rev = new_reverb();
-    init_stereo_reverb(amy_global.bus[bus]->reverb.rev);
+    return init_stereo_reverb(amy_global.bus[bus]->reverb.rev);
 }
 
 void dealloc_reverb_delay_lines(uint8_t bus) {
@@ -352,7 +352,10 @@ void config_reverb(uint8_t bus, float level, float liveness, float damping, floa
         //printf("config_reverb: level %f liveness %f xover %f damping %f\n",
         //      level, liveness, xover_hz, damping);
         if (amy_global.bus[bus]->reverb.level == 0) {
-            alloc_reverb_delay_lines(bus);
+            if (!alloc_reverb_delay_lines(bus)) {
+                amy_global.bus[bus]->reverb.level = 0;
+                return;
+            }
         }
         config_stereo_reverb(amy_global.bus[bus]->reverb.rev, liveness, xover_hz, damping);
     }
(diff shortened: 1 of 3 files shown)


bwhitman:
### 🎛️ AMYboard HW CI

bwhitman:
### 🎛️ HW CI (physical bench)

**[⬇️ Artifacts: recordings · screenshot · serial logs · run logs](https://github.com/shorepine/tulipcc/actions/runs/27889688769)**

PR #750, opened 2026-06-22
Issue #720: Rework zero-amp termination - now only in release.  Skip …

dpwe:
…oscs where amp[0] == 0. Change last_amp.

Issue #720 responded to a couple of instances where voices with very slow attacks were never being started because of some logic that rapidly terminates new oscs that don't appear to be emitting sound, in order to avoid wasted computation on Juno patches.  Instead, we now skip rendering oscs whose `amp[CONST]==0` (we already skipped them if their `msynth.amp` and `msynth.last_amp` were zero, but this is more aggressive and avoids running `hold_and_modify`).

In testing this, I noticed that some AM envelopes were very noisy because we had `msynth.last_amp` logic that failed to linearly interpolate rising AM envelopes.  I forget why I put this in, I think it was to do with voice-stealing clicks, but anyway I simplified it and most tests were unchanged.  I listened to all the changed tests (with the new `make playfailed`) and they all sounded OK.


bwhitman:
### 🎛️ AMYboard HW CI

dpwe:
I tested this build against the previous  version for the number of simultaneous Juno 1 voices that can be rendered on an RP2350 (4), and it's surely no worse than it was before.

bwhitman:
### 🎛️ HW CI (physical bench)

**[⬇️ Artifacts: recordings · screenshot · serial logs · run logs](https://github.com/shorepine/tulipcc/actions/runs/27923038172)**

PR #753, opened 2026-06-22
Add Arduino example showing how to implement custom audio effects in AMY

strangebroadcasts:
Adds an example sketch showing how AMY's `amy_get_input_buffer` and `amy_set_external_input_buffer` can be used to implement custom DSP effects (in this case, a simple bit reduction/bitcrusher)


bwhitman:
This is great, thank you @strangebroadcasts !

PR #756, opened 2026-06-23
src/patches.c: Don't ignore FX commands when synth is specified (issue #755).

### LINKED ISSUE #755: FX commands are ignored when directed to a synth
I noticed that Juno patches were no longer using chorus on the AMYboard.  The AMY tests, however, were passing, including many voices with chorus.

The difference is that AMYboard was not loading the Juno patches directly (with something equivalent to `amy.send(synth=1, patch=0)`), but instead read the Juno patches, reinterpret them for the two-osc "K257" patch, then send them as individual wire commands, including commands for the global FX.  These FX are prefixed with, e.g., `i1` to direct them to synth 1.

Executing AMY events with `synth` set vector through `patches_event_has_voices()`.  In #750, I added an early-exit to that routine to skip events that were effectively empty because they involved only patch changes and had already been handled by `patches_load_patch()`.  Unfortunately, it appears that patch-directed global FX events fell through the cracks here, being events that should not be sent to all oscs (which was the purpose of the early-exit), but which had *not* been handled by `patches_load_patch`.

The solution is to spot these events in the early-exit code and handle appropriately.

dpwe:
Fixes #755


bwhitman:
### 🎛️ AMYboard HW CI

bwhitman:
### 🎛️ HW CI (physical bench)

**[⬇️ Artifacts: recordings · screenshot · serial logs · run logs](https://github.com/shorepine/tulipcc/actions/runs/28029528487)**

PR #759, opened 2026-06-23
Fix MIDI Program Change silencing AMYboard Web Editor patches (#758)

### LINKED ISSUE #758: MIDI Program Change on top of AMYboard Web Editor patches
AMYboard users report that MIDI PC makes their AMYboard go silent.

MIDI PC (Program Change) changes patches via a "program number" in a 7 bit range.  This range is extended by MIDI Control Code 0, Bank Select, which selects a page of 128 programs.

AMY has 258 built-in patches; 0-127 are Juno 6, 128-255 are DX7, 256 is the additive piano, and 257 is the 2-osc AMYboard Web Editor base patch.

When AMY receives a PC but has never seen a Bank Select, it uses bits 8 upwards of the current patch number as the bank.  So, PC on a Juno patch selects another Juno patch, and on a DX7 patch it selects another DX7 patch.

Unfortunately, this means that on an AMYboard Web Editor patch, it selects some other patch in the 256-383 range, most of which are undefined.  I think this is when the board goes silent.

Quick fix would be to map "inferred bank 2" back to zero.

Better fix would be to unify the AMYboard Web Editor and the built-in Juno patches, so you were already in bank 0.  But the AMYboard Web Editor can't properly emulate Juno patches that use more than 2 of (pulse, saw, subosc, noise).  Not clear how big a problem this is, since we already expose this imperfect emulation in the AMYboard Web Editor, and no-one seems to have objected.

bwhitman:
Fixes #758.

## Problem

AMYboard users reported that sending a MIDI Program Change makes the board go silent.

When AMY receives a PC but has never seen a Bank Select (CC 0), `amy_received_program_change` infers the bank from the high bits of the *current* patch number, so a PC on a Juno voice stays in the Juno bank and a DX7 voice stays in DX7:

```c
bank_number = (instrument_get_patch_number(e.synth) & 0xFF80) >> 7;
```

The built-in patch tables hold 258 patches: `0-127` Juno, `128-255` DX7, `256` additive piano, `257` the AMYboard Web Editor base patch. Patches `256+` infer **bank 2**, so a PC maps to `program + 256` — the `258-383` range, which runs past the end of the 258-entry `patch_commands` / `patch_oscs` tables. That out-of-bounds read yields undefined oscillators: silence on the board (and, it turns out, a **segfault** in the desktop test harness).

## Fix

Only banks 0 (Juno, `0-127`) and 1 (DX7, `128-255`) are full 128-patch banks. Clamp the *inferred* bank so anything higher falls back to bank 0 (Juno) — this is the quick fix DAn suggested in the issue:

```c
bank_number = (instrument_get_patch_number(e.synth) & 0xFF80) >> 7;
if (bank_number > 1) bank_number = 0;
```

Explicit Bank Selects (CC 0) are untouched — this only affects the inferred-bank path.

## Testing

- Added `TestProgChangeOnPreset257`: puts synth 1 on patch 257, fires a Program Change, and checks the note lands on Juno patch 5 and sounds. With the clamp removed this test **segfaults**; with the fix it passes bit-exact.
- Full suite: **99 tests pass**.

## Follow-up (not in this PR)

The issue also floats a "better fix": unify the Web Editor base patch with the built-in Juno patches so a 257 synth is already in bank 0. That's a larger change (the Web Editor can't perfectly emulate Juno patches using >2 of pulse/saw/subosc/noise), so this PR is just the targeted quick fix.


bwhitman:
### 🎛️ AMYboard HW CI

bwhitman:
### 🎛️ HW CI (physical bench)

**[⬇️ Artifacts: recordings · screenshot · serial logs · run logs](https://github.com/shorepine/tulipcc/actions/runs/28035433021)**

PR #780, opened 2026-06-29
Fix #779: AMYboard chord crackle under esp32 core 3.3.9+ (amp-path transcendentals + -O2)

### LINKED ISSUE #779: BillieJeanScheduled has dropouts on AMYboard
The `BillieJeanScheduled` Arduino example script (with drums, bass, and chords) has some audible dropouts on AMYboard when the chords come in. The same code plays cleanly on an ESP32 dev board with an external PCM5101 DAC.

I went back to AMY Arduino 1.2.6 and the problem persisted.

I went back to AMY Arduino 1.2.4 and it was OK.

(`BillieJeanScheduled` will not build under 1.2.5 because of the change to the `config_reverb` signature.)

Something about the AMYboard setup is making less CPU available, and it has become visible lately.

AMY v1.2.4 was released 2026-03-24

AMY v1.2.6 was released 2026-06-04

There were a ton of changes in-between.

There were some changes to the RTOS scheduling priorities.  I tried reverting those, it wasn't that.

bwhitman:
Fixes #779.

## Symptom
`BillieJeanScheduled` crackles on AMYboard when the chords come in — but only under a newer Arduino-esp32 core. Same AMY build:

| AMY | core 3.3.8 | core 3.3.9 / 3.3.10 |
|-----|-----------|---------------------|
| 1.2.4 | clean | clean |
| 1.2.6 → 1.2.17 | clean | **crackles** |

So it's a latent AMY regression that only crosses the timing margin on the tighter 3.3.9+ core (both cores are on ESP-IDF v5.5.4; the delta is the arduino wrapper). A first ear-bisect on 3.3.8 found nothing because both endpoints are clean there — re-bisecting on 3.3.9 (where 1.2.4=clean / 1.2.6=crackle) isolated it cleanly.

## Root cause
First bad commit: `61a68322` ("First start of log-domain amp calculation"). It moved the per-oscillator amp calc in `hold_and_modify()` from a multiply-accumulate to a log-domain map, so `amp_combine_controls()` / `map_60dB_to_01f()` now run **a `powf(10,x)` plus up to ~8 `log2f` calls per audible oscillator, per render block**. A 6-voice chord (+ bass + drums) multiplies that into a render overrun on the 3.3.9 core. (Not a DMA/I2S issue — purely CPU in the hot path.)

## Fixes (two independent levers, each validated on AMYboard + core 3.3.9)

**1. Cut the transcendentals in `amp_combine_controls()`**
- Skip coefs that are `0` — `0 * val == 0`, so it's **bit-exact** and removes most `log2f` calls (typical oscs only have const/vel/eg0 nonzero).
- Replace libm `powf(10,x)` → `exp2_lut(x·log2 10)` and `log2f` → `log2_lut` (AMY's existing fast LUTs, already used in `envelope.c`).
- Clears the crackle **at stock `-Os`**.
- ⚠️ The LUT swap is **not bit-exact**: the CPython WAV suite shows worst-case err ≈ **−92 dB** (~50–70 dB below signal, i.e. inaudible). **62 golden refs were regenerated.** @dpwe — please confirm you're happy with this precision tradeoff on the amp curve; happy to drop just the LUT half and keep the bit-exact zero-skip if you'd rather.

**2. Build AMY's DSP hot path at `-O2`**
- `#pragma GCC optimize("O2")` in `amy.h` for `ESP_PLATFORM` compiles (dpwe modified).
- The Arduino esp32 core builds libraries at `-Os` and the AMYboard board exposes no optimization menu, so AMY self-elevates — **no arduino-esp32 PR required**, and it helps every CPU-bound AMY sketch.
- `-O2` not `-O3` because arduino-pico's `-O3` has crashed RP2350.
- On its own, `-O2` also clears the crackle with the **stock** amp code — so the two fixes are belt-and-suspenders.

## Testing
- CPython suite: **100/100 pass** (`make test`) against the regenerated goldens.
- Hardware: AMYboard + esp32 core 3.3.9, `BillieJeanScheduled` — confirmed clean with the LUT fix alone (`-Os`), with the `-O2` pragma alone (stock amp), and with both.


bwhitman:
### 🎛️ AMYboard HW CI

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1061

bwhitman:
### 🎛️ HW CI (physical bench)

**[⬇️ Artifacts: recordings · screenshot · serial logs · run logs](https://github.com/shorepine/tulipcc/actions/runs/28414129706)**

PR #787, opened 2026-07-03
Fix stereo_reverb LPF state never being written back

rt-rtos:
LPF() in src/delay.c implements the reverb feedback crossover lowpass, but takes its state parameter by value - the state += update modifies a local copy and is discarded. The f1state..f4state fields in `reverb_params_t` are read at the four call sites in `stereo_reverb()` but never written, so they stay at 0 forever and the "1-pole lowpass" degenerates into a memoryless gain on the instantaneous sample.

As a result, the reverb feedback path has no lowpass at all, and `config_stereo_reverb()`'s crossover_hz / damping parameters act as level trims rather than spectral controls - the Stautner-Puckette design (and the rev2~ patch referenced in the comments) filters the feedback to darken the tail. Both fixed-point and float builds are affected, since the bug is in parameter passing, not arithmetic.

Fix: LPF() takes `SAMPLE *state` and updates it in place; the four call sites pass `&rev->fNstate.`

Verified by rendering a noise burst through the reverb (level 1, liveness 0.85, damping 0.5, crossover 3000) and tracking the high-frequency content of the tail: before, the tail spectrum stays flat until it decays into noise; after, HF content falls monotonically pass by pass. With the fix, crossover_hz behaves as expected (1 kHz darkens much faster than 3 kHz; 12 kHz ≈ no filtering) and damping 0 cleanly disables the effect.

Note: this is an audible behaviour change by design. Tails become darker and somewhat longer at the same settings, since damping no longer doubles as a tail-level trim. Patches tuned against the broken behaviour will sound different.


rt-rtos:
The CI failures on TestBuses and TestSynthBusCmds are expected: they're the only two tests that enable reverb, and their reference WAVs were rendered with the old unfiltered feedback path. Since this fix intentionally changes the reverb output (the tail is now actually lowpassed, so it's darker and decays more naturally), those refs can no longer match sample-for-sample — the −36 dB error is the audible change itself, not a regression. All 98 non-reverb tests still pass at the CI threshold. The follow-up commit regenerates just those two reference WAVs from the fixed build; with them the full suite passes

dpwe:
Awesome. This explains why your earlier fix to the reverb lpf calculation *didn't* affect the tests. Thank you!

rt-rtos:
I will finish a PR tomorrow that touches the same surface and would conflict until this lands. That should offset the four extra trips we take to memory and more by caching reverb state locally as we do with `delay_line_in_out`

dpwe:
This PR was duplicated in #809, so this (original) version has no need to be merged.

PR #790, opened 2026-07-04
 Hoist stereo_reverb() delay-line state into locals

rt-rtos:
stereo_reverb() interleaves ten delay lines (six early-reflection taps + four feedback lines) per sample, and each DEL_IN/DEL_OUT call re-reads that line's samples pointer, len, next_in, and fixed_delay from the struct on every access. That's redundant traffic on a hot per-sample loop, and on platforms where delay lines live in external RAM the struct header itself is slow to touch.

This hoists each line's state into locals at loop entry (DL_LOAD), does all reads/writes through those locals (DL_READ/DL_WRITE), and writes the advanced next_in values - plus the four fNstate LPF states- back to the struct once after the loop. next_in is the only field that changes inside the loop, so it's the only one that needs writing back; samples, len, and fixed_delay are read-only for the duration of the call.

No behavior change, same read/write order and arithmetic per sample, just fewer memory accesses to get there. Verified against the existing test suite (bit-identical output) and inspected asm output on xtensa-esp32s3-gcc the per-sample loop body drops from 327 to 290 instructions (-11%), and loads from 96 to 85. The whole-function instruction count goes up slightly (437 → 509) from the one-time prologue/epilogue and register spills for the extra locals, but that's a fixed cost that amortizes over AMY_BLOCK_SIZE (128/256) samples per call.

Stacked on #787 (fix-reverb-no-lowpass) — this diff only makes sense on top of that fix, since the fNstate write-back only exists once LPF() takes its state by pointer. 

Edit: Comment on lines 288 - 304 is more verbose than is the norm for this codebase. I deemed this to be worth it because the preprocessor magic going on is not very intuitive and is hard to follow. Tell me if you find this too bloated and want me to write a shorter summary.  You are also free to edit it as you see fit.


dpwe:
I cloned this into #811 so I could fix the conflicts and merge.  I love this optimization!

PR #793, opened 2026-07-04
Fix LPF24 fixed-point blow-up when the state outgrows the block-start headroom bound (patch 55, vel 0.51–0.57)

bwhitman:
## Symptom

```python
amy.send(synth=1, patch=55, num_voices=4)   # Juno A78 Wah Release
amy.send(synth=1, note=52, vel=0.55)
```

glitches (output blows up to full scale and hits the clipper) near the end of the ~11 s decay — but **only for velocities between ~0.51 and 0.57**, not above or below. Patch 55 has the VCF envelope negated (`F…,-1.646,…` with `R4.78`), so the cutoff sweeps *up* through the note's harmonics as the amplitude decays.

## Root cause

In `dsps_biquad_f32_ansi_split_fb_twice()` (the `FILTER_LPF24` path; `_once` has the same pattern), the pseudo-float multiply shift counts are precomputed **once per block**:

```c
bbits = nheadroom16(max_val);
cbits = nheadroom16(SHIFTL(MAX(state_max, max_val), 2));   // 2 bits of growth allowance
```

`cbits` is then trusted by `top16SMUL_after_a()` for every state-side multiply in the loop (`e·y1`, `f·y2`, `e·v1`, `f·v2`, and `a·w0` where `w0 = v0 + 2v1 + v2` is already ~4× the per-stage state). The hidden assumption is that the state can't grow more than ~4× within one block.

A downward-tracking VCF respects that: the resonant ring-up builds while the state is already large, so `state_max` tracks it. An **upward** sweep does not — the state is small (the cutoff was below the harmonics), then the cutoff steps onto them and two cascaded Q≈4.8 sections ring up ~Q² ≈ 23× within a block or two. The under-shifted operand keeps more than 16 significant bits, the 32-bit product wraps, and the filter explodes. (There's an old fingerprint of the same failure mode in `filter_process()`: `normbits = MIN(8, normbits);  // Without this, I get a weird sign flip at the end of TestLFO - intermediate overflow?`)

**Why only vel 0.51–0.57:** the shift counts quantize to whole bits, so the effective overflow ceiling is 16–32× the block-start max depending on where that max sits inside its power-of-two range — and ~23× lands right between. Velocity slides the (decaying) amplitude trajectory against the velocity-independent cutoff trajectory; only a narrow band puts the critical block's max in the losing sliver. One notch louder → one more bit dropped → safe; quieter → the product stays under 2³¹ → safe.

## Fix

Measure the state operand's **actual** headroom at each multiply (`top16SMUL_after_a_live()`, i.e. what plain `top16SMUL()` already does for `b`) instead of trusting the block-start bound. Cost is one CLZ per state multiply (`NSAU` on Xtensa, `CLZ` on ARM); the coefficient-side preprocessing (`top16SMUL_a_part`) is unchanged. `state_max`/`cbits` go away. As a bonus, when the state is *below* the old bound the live measurement drops fewer bits, so state precision improves slightly.

## Verification

Rendered the repro above for 12 s across velocities (fixed-point build, offline `amy.render`), peak |sample|:

| vel | before | after |
|---|---|---|
| 0.50 | 0.387 | 0.387 |
| 0.51 | **1.000 (clipped)** | 0.395 |
| 0.53 | **1.000 (clipped)** | 0.410 |
| 0.55 | **1.000 (clipped)** | 0.426 |
| 0.57 | **1.000 (clipped)** | 0.441 |
| 0.58 | 0.449 | 0.449 |

After the fix the peak is exactly linear in velocity through the whole window, as a linear filter should be.

`make test`: baseline (unmodified clone) was 100/100 bit-exact in my environment. With the fix, 32 filter-touching tests move by LSBs — all at err ≤ **−83 dB** except `TestJunoClip` (−33 dB), where four loud voices sit on the clipper boundary that amplifies LSB differences. Second commit regenerates those refs → 100/100 pass. Happy to drop that commit if you'd rather regenerate/listen yourself.

I didn't add a regression test: the blow-up needs the full ~11 s envelope alignment, and a time-compressed version of the patch doesn't hit the same bit alignments within the 1 s test render. The repro script is short if you want to check it on your end:

```python
import numpy as np, amy, c_amy as _amy
for vel in (0.50, 0.55, 0.58):
    _amy.stop(); _amy.start(0)
    amy.send(time=0, synth=1, patch=55, num_voices=4)
    amy.send(time=50, synth=1, note=52, vel=vel)
    print(vel, np.max(np.abs(amy.render(12.0))))
```


bwhitman:
### 🎛️ AMYboard HW CI

bwhitman:
Heads up: the ❌ HW CI comment above was a **bench collision, not a firmware failure** — I was running manual A/B glitch sweeps on the same physical bench when the CI run started, and my `arecord` was holding the capture device (hwci died with `FileNotFoundError` on its recording temp file). The re-dispatched run should be clean; I'm staying off the bench until it finishes.

Also verified locally at 1d7c1d7f (fixed-point desktop build): the patch-32 `note=71` note-off burst is gone (max|x| 0.569, 0 clipped samples — was full-scale with 312 clipped samples before the slew tightening; the burst was ~14.5× the note level and scales as a *linear* transient, reproducible in a float build too, so the 3.0/Q slew limit is the right kind of fix). Patch-55 wah stays linear through the old 0.51–0.57 window.

bwhitman:
### 🎛️ HW CI (physical bench)

**AMYboard** (USB-MIDI + AMY `zP` → audio; built-in tones + AMYboard World sketches acid/house/universal_hair/woodpiano over the SysEx control API): ✅ **PASS** — flashed this PR’s firmware; all checks matched the references.

**[⬇️ Artifacts: recordings · screenshot · serial logs · run logs](https://github.com/shorepine/tulipcc/actions/runs/28718422582)**

<sub>Self-hosted bench. Audio spectral-compared to `ref/hwci_basic.wav`, the AMYboard World sketch refs (`ref/{acid_generator,house_generator,universal_hair,woodpiano}.wav`) + `ref/tulip_basic.wav`; Tulip screenshot pixel-compared to `ref/tulip_screenshot.png`. Both analog outs share one capture card, so the tests run sequentially.</sub>

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1079

PR #797, opened 2026-07-05
Gamma9001: fill TR-808 gaps and add the 80s Power Kit bank [skip release]

bwhitman:
Two sample-bank additions to the Gamma9001 drum machine (docs-only, hence `[skip release]`):

## TR-808 gap fill
The Koblo Tokyo distribution never shipped an 808 snare (or rimshot, toms, cymbal). `extract_sf2_extras.py` pulls the seven missing sounds from `sounds/HS-TR-808-Drums.sf2` — already in the repo, already the source of AMY's baked-in ROM kit. Adds TR-808 Snare 1/2/3 (SNR 4, 7, 12), Rimshot, Tom Lo/Hi, Cymbal — the 808 bank now covers a full GM core kit.

## 80s Power Kit
A user asked for the GS "Power set" — the gated-reverb 80s rock kit. `extract_power_kit.py` parses the Korg AG-10 SoundFont's Power Kit preset (bank 128, program 16 — the GS-standard slot) through its preset→instrument→zone chain, so each sample arrives with its GM note assignment: real/metal kicks, power + gated snares, six processed toms, hats, cymbals, clap, cowbell. 20 samples, silence-trimmed to 1.9 MB. A "Power Ballad" example song in the dropdown shows it off.

Also reorganizes the sound picker into one scrollable list with a section per bank (in kit order) plus an "Other" bucket at the bottom for the non-bank samples; the top buttons jump to sections.

Both extractors are idempotent and append manifest entries at the end, keeping preset indices in saved/shared setups stable. Verified in the browser: 155 samples load, both new banks play, the example song runs.

Note: the AG-10 SoundFont itself is not committed — only the 20 extracted drum one-shots.


bwhitman:
Consolidated into #798, which now targets main and contains all of this branch's commits.

PR #804, opened 2026-07-05
Fix start_sample capturing at parse time instead of scheduled event time

bwhitman:
## Bug

`amy.start_sample(preset=1024, source=amy.SAMPLE_FROM_OUTPUT, ...)` followed by PCM playback of that preset played back low-frequency "garbage" even when the intended capture window was bit-exact silence.

Root cause: the `zS` (`start_sample`) and `zO` (`stop_sample`) wire commands executed directly inside the parser, bypassing the delta queue. On any scheduled timeline (offline render, `time=`-sequenced messages), all messages parse before rendering begins, so the capture always armed at render block 0 — grabbing the **start of the render** (the earlier synth note) instead of the intended window. The "50–100 Hz LF junk" in the repro was the Juno patch's sub-oscillator (131 Hz, C3), replayed at `vel=16`.

Two adjacent latent bugs are also fixed:
- `pcm_load` never zeroed sample RAM, so a capture stopped before `max_frames` (via `stop_sample`) played uninitialized heap for the uncaptured remainder.
- The `zS` handler hardcoded `channels=2` where the capture path sizes by `AMY_NCHANS`.

## Fix

- New `START_SAMPLE` / `STOP_SAMPLE` delta params: `zS`/`zO` now schedule through the normal delta queue and fire at event time (`src/parse.c`, `src/amy.c`, `src/amy.h`). The preset RAM is still allocated (and now zeroed) at parse time; `pcm_arm_sample_capture()` (`src/pcm.c`) arms the capture when the delta fires. Untimed messages behave as before — capture starts on the next rendered block.
- `amy.start_sample()` / `amy.stop_sample()` gain an optional `time=` kwarg (no `_KW_MAP_LIST` change, so no generated-API regen needed).
- Docs updated in `docs/synth.md`.

## Test

New `TestSampleSilence` regression test: renders a sine note, captures a verified bit-silent window after it decays, plays the capture back at `vel=16`, and — in addition to the ref-waveform comparison — explicitly asserts the playback region is all-zero. With the fix reverted it fails with `PLAYBACK NOT SILENT (peak 2784 LSB)`; with the fix, the full suite passes 103/103. The pre-existing `TestSample` ref is unchanged, and scheduled/untimed non-silent captures round-trip with aligned cross-correlation 1.000 against the source window.


bwhitman:
### 🎛️ AMYboard HW CI

bwhitman:
### 🎛️ HW CI (physical bench)

**AMYboard** (USB-MIDI + AMY `zP` → audio; built-in tones + AMYboard World sketches acid/house/woodpiano over the SysEx control API): ❌ **FAIL** — a check did not match, or the run errored. See the log/artifacts.

**[⬇️ Artifacts: recordings · screenshot · serial logs · run logs](https://github.com/shorepine/tulipcc/actions/runs/28757518843)**

<sub>Self-hosted bench. Audio spectral-compared to `ref/hwci_basic.wav`, the AMYboard World sketch refs (`ref/{acid_generator,house_generator,woodpiano}.wav`) + `ref/tulip_basic.wav`; Tulip screenshot pixel-compared to `ref/tulip_screenshot.png`. Both analog outs share one capture card, so the tests run sequentially.</sub>

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1094

Issue #808, opened 2026-07-06
Fix `pan=` etc for `SYNTH_NOTES_VIA_MIDI` note events


dpwe:
GM Drums (channel 10) are handled by rewriting the different drum notes via `midi_note_cmd` in #777, including allowing AMY-generated events to go through the same mapping when `synth_flags=amy.SYNTH_FLAGS_NOTES_VIA_MIDI` is set.

This works fine to allow commands like `KICK=35; amy.send(synth=10, note=KICK, vel=1)`.  However, arguments other than `note` and `vel` in such events are ignored - the event is being turned into a MIDI note-on, which does not carry other parameters.

Instead, we would like for commands like `Amy.send(synth=10, note=KICK, vel=1, pan=0)` to work as expected (kick drum in the left ear).  I think we can do this by restructuring `midi_message_handler` to yield a series of `amy_events` rather than directly dispatching them via `Amy_add_message`.  Then, when `patches_event_has_voices` wants to perform MIDI mapping, it can set additional fields in the event structures before feeding them into the delta queue.

PR #845, opened 2026-07-08
Store sequenced notes on NOTES_VIA_MIDI synths instead of playing them at add time

bwhitman:
## Problem

A sequenced note event on a synth with `SYNTH_FLAGS_NOTES_VIA_MIDI` — e.g. the GM drum kits, `amy.send(synth=10, patch=384, num_voices=6, synth_flags=3)` per the docs — was routed through `midi_msg_handler` → `amy_add_message`, which ignores the delta queue the caller was building. For `sequencer_add_event` that meant:

- the note played **once, immediately, at the moment the sequence was defined** (even if scheduled seconds in the future), and
- **nothing was stored** in the sequence entry.

So a sequenced drum pattern (e.g. Tulip/AMYboard `sequencer.AMYSequence` sketches) sounded like a single composite hit at load, then went silent forever, while sequenced melodic synths on the same setup kept playing. Regressed with the move of GM drum translation to MIDI note-commands (fac0d21 / cc581f4); reported via an AMYboard generated-sketch "drums only play one note" bug.

## Fix

`midi_msg_handler_to_queue()` carries the caller's target queue through the MIDI note-command mapping:

- **Storing** (queue is a sequence entry or patch definition): the mapped message is parsed with `yield_event_from_message` and converted onto that queue via `amy_event_to_deltas_queue`. The existing `iM<chan>` marker prevents the events re-entering the mapping when they resolve to voices.
- **Live** (queue is NULL or the global delta queue): unchanged — still `amy_add_message`, preserving time/latency scheduling semantics.

`midi_msg_handler` is now a shim over the new function; `patches_event_has_voices` passes its queue through.

## Test

`TestSequencedSynthDrums`: schedules a kick on default synth 10 far in the future and asserts the next second of rendering is silent. Fails at **-24.9 dB** (drum sounding at add time) before the fix; passes after. Full suite: 104 tests pass.

Verified end-to-end with the original failing AMYboard sketch pattern (32-step 808 pattern on `synth=10, patch=384, synth_flags=3`): pre-fix it collapses to one hit then silence; post-fix it loops correctly, including with `num_voices=1`.

Related: #808 — same mechanism drops `pan=` etc. from NOTES_VIA_MIDI note events; the restructuring proposed there (yield events, then decorate) would subsume this, but this is a minimal fix for the sequencer-losing-drums regression.

### DIFF
--- a/amy/test.py
+++ b/amy/test.py
@@ -934,26 +934,6 @@ class TestSynthDrums(AmyTest):
     amy.send(time=900, synth=10, note=37, vel=0)  # snare note off - ignored with current setup.
 
 
-class TestSequencedSynthDrums(AmyTest):
-  """A sequenced note on a NOTES_VIA_MIDI synth (GM drums) is stored for the
-  sequencer, not played immediately at add time."""
-
-  def test(self):
-    _amy.stop()
-    _amy.start(1)  # default synths: channel 10 is the GM drum synth.
-    amy.render(0.5)  # Let the default synth setup settle.
-    # Schedule a kick far in the future (tick 50*PPQ ~ 30s away) on a looping
-    # sequence.  Nothing should sound during the next second of rendering.
-    amy.send(sequence='%d,%d,0' % (50 * amy.AMY_SEQUENCER_PPQ, 100 * amy.AMY_SEQUENCER_PPQ),
-             synth=10, note=35, vel=1)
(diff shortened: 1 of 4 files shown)


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1117

PR #848, opened 2026-07-08
interp_partials: keep pitch-table reads in bounds for out-of-range notes (fix OOB read)

bwhitman:
## Summary

`interp_partials_note_on()` clips velocity to the preset's table range but never bounded **pitch**. The piano partials table (`piano_notes[]` in `src/interp_partials.h`) covers MIDI 24..104. For any note above 104, the pitch-index search loop walks `pitch_index` to `num_pitches - 1`, and then every `pitch_index + 1` access — `pitches[pitch_index + 1]`, `num_harmonics[note_number + num_velocities + 1]`, and the `_harmonic_base_index_for_pitch_vel(pitch_index + 1, ...)` harmonics rows — reads past the end of the tables.

Empirically (host build at ae5a1d8fe): playing chords of notes 108/112/115 on the piano patch while other notes are held, repeated a few times, **permanently silences all AMY output** (even unrelated plain-sine oscs). Re-creating the synth recovers it. This is the piano wedge seen in tulipcc with `patch=256`.

## Fix

Per @dpwe's review, out-of-range notes are **extrapolated, not clipped**: the pitch-index search loop is bounded at `num_pitches - 2` so `pitch_index + 1` always stays inside the tables, and `midi_note` itself is left untouched. Notes outside the table range get a `pitch_alpha` outside [0, 1], i.e. linear extrapolation from the two edge rows of the table — so they still sound (if wacky) instead of pinning to the edge pitch. Velocity keeps its existing clip, which also keeps `vel_index + 1` in bounds.

## Test

New `TestInterpPartialsOutOfRange` regression test: piano synth with 6 voices, four rounds of an in-range note held while an out-of-range chord (108/112/115) plays, then a final in-range note that must still sound. Verified the test's event pattern reproduces the wedge on the unfixed build (final note renders pure silence); reference regenerated for the extrapolation behavior.

- `make test`: **104 tests pass** (was 103 + this new one)
- The original 40-iteration wedge repro runs clean on the fixed build (`died_at: None`), and unrelated oscs stay alive.


dpwe:
It's nice to be able to extrapolate pitch (if wacky).  Instead of clipping note value, we ensure pitch table reads stay within bounds.

bwhitman:
Updated for the extrapolation approach: fixed the now-stale comment above the velocity clip (it still described a pitch clip), verified `make test` passes all 104 tests with the regenerated `TestInterpPartialsOutOfRange` reference, and refreshed the PR title/description to match.

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1122

PR #850, opened 2026-07-09
midi_message_handler now layers onto a base_event, can propagate other params.

### LINKED ISSUE #808: Fix `pan=` etc for `SYNTH_NOTES_VIA_MIDI` note events

dpwe:
Fixes #808


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1119

PR #854, opened 2026-07-09
Sequencer: tick from the render loop, not per-platform timers/threads

### LINKED ISSUE #849: Fix scheduler triggering to work in tests/non-real-time rendering
Right now, the sequencer is updated by calls to `sequencer_check_and_fill()`, which will copy the stored delta strings to the global delta queue when the time is right.  `sequencer_check_and_fill()` is called by a 0.5 ms periodic timer on MCUs (ESP32, RP2040), by a real-time POSIX thread on desktop, and by a hook in mini audio, I think for emscripten/web.

Currently, the scheduler doesn't work under tests (usually, no events occur).  I think this is because the rendering in the test harness is actually much faster than real-time, and it's certainly not in-sync with real-time, so the POSIX mechanism wouldn't work.

Instead, I think it's legit for the scheduler to work on "AMY time", as counted by the number of audio samples generated.  I think we can move the `sequencer_check_and_fill()` call to the top of `amy_render()` (the once-per-AMY-block call to compute samples) and stuff will happen at the same time anyway (since nothing happens until the deltas are processed).  However, we don't have any automated tests for this, so we need to take care when making this change.

Win would be (a) simplification of the scheduler - no per-platform specialization, handled uniformly, and (b) scheduler would work under `test.py` (and other non-real-time rendering contexts).

bwhitman:
Fixes #849.

`sequencer_check_and_fill()` is now called at the top of `amy_execute_deltas()`, once per rendered block, and all the per-platform timer/thread pollers are deleted (esp_timer, pico repeating timer, POSIX thread, Daisy TIM5, plus the `_sequencer_start/_stop` scaffolding). Since `amy_sysclock()` counts rendered samples, those pollers added no timing precision — ticks were always decided against the sample clock, and the fired deltas are block-quantized anyway. Net effect: the sequencer now works identically in real-time and non-real-time rendering, including under `test.py`.

**Web:** the render loop runs in the AudioWorklet thread, where `EM_ASM` can't reach the page's `amy_sequencer_js_hook`. The tick `EM_ASM` moves out of `sequencer_process_tick()` into a new `sequencer_check_and_call_js_hook()`, called from the emscripten main loop on the browser main thread. It replays each elapsed tick to the hook, with a one-second skip-ahead cap mirroring `sequencer_check_and_fill()`'s catch-up guard. The hook contract is unchanged for tutorial.html / piano.html / Tulip web (`mp.tulipTick`) — it was already rAF-burst driven. A side benefit: sequences now keep playing when the tab is backgrounded, since ticking no longer depends on rAF.

**Tulip native:** `amy_external_sequencer_hook` is now invoked from the render context instead of a timer; `tulip_amy_sequencer_hook` only calls `mp_sched_schedule`, which is safe there.

**Two latent restart bugs surfaced by the new test** (both predate this change, but were masked because the sequencer never ran under tests):

1. `oscs_init()` reset `total_blocks` *after* `sequencer_init()` had anchored `next_amy_tick_us` against the previous session's sysclock — so after any `amy_stop()`/`amy_start()`, the first tick was scheduled a whole session-length into the future. The clock reset moves to the top of `oscs_init()`.
2. `sequencer_running` / `sequencer_external_clock` are process-lifetime statics and survived restarts; `sequencer_init()` now resets them to boot state (running, internal clock).

**Test:** new `TestSequencer` covers absolute-tick events (note on at tick 20, off at tick 40) and a periodic event (every 60 ticks), rendered offline. The previous `tests/ref/TestSequencer.wav` golden was silence; it's regenerated with the real output. All 104 tests pass, and the 103 pre-existing goldens are bit-exact — the change doesn't alter any rendered audio outside the sequencer.

Verified: native build (`make`), web build (`make web`), full test suite twice (deterministic), and a poisoned-restart repro (MIDI inject → stop/start → sequenced events still fire).


bwhitman:
Rebased onto main and resolved the conflict with the sequencer tests Dan added in the meantime:
- Kept `TestSequencer` and `TestSequencedSynthDrums`, minus the "DOES NOT WORK, see #849" notes — they make sound now.
- `TestSequencedSynthDrums` now runs the line Dan originally wanted (`sequence` on the NOTES_VIA_MIDI synth 10) instead of the simulation workaround — it works with the sequencer ticking in the render loop.
- My osc-level test is renamed `TestSequencerOsc`.
- Regenerated all three goldens (they were silence). 107 tests pass, deterministic across two runs.

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1121

PR #864, opened 2026-07-10
Add MIDI clock OUT: external_midi_sync=2 makes AMY the realtime clock master

bwhitman:
## What

AMY could already *follow* MIDI realtime clock (`external_midi_sync=1`, wire `zC1`) but had no way to *send* it. This PR extends the same switch with a third mode:

- `0` (default): internal clock; realtime messages neither followed nor sent — unchanged
- `1`: follow incoming `0xF8`/`0xFA`/`0xFC` — unchanged
- `2` (**new**): AMY is the clock master — sends `0xF8` at 24 PPQ derived from AMY's own `tempo` (one pulse per two 48 PPQ sequencer ticks), plus `0xFA`/`0xFC` when the sequencer transport starts/stops (`zY1`/`zY0`, `sequencer_run=`, or the C API)

Per the MIDI spec recommendation, clock keeps flowing while the transport is stopped, so downstream slaves stay tempo-locked and start cleanly on the next `0xFA` (the first `0xF8` after Start marks the downbeat; the phase counter resets on Start).

Usable from C (`amy_external_midi_sync(2)`), Python (`amy.send(external_midi_sync=2)`), the wire/sysex (`zC2`), and JS — no `_KW_MAP_LIST` change needed since the kwarg already existed.

## How

- Emission lives in `src/amy_midi.c` (`midi_clock_out_tick/start/stop/enabled`), called from the sequencer's internal tick path. In follow or off mode they're no-ops, so the external-clock path can't echo clock back out.
- `sequencer_check_and_fill()` now keeps the tick timer running while the transport is stopped **only** when clock-out is enabled, emitting clock without processing sequence events; behavior in modes 0/1 is unchanged.
- **Web:** the sequencer ticks in the AudioWorklet thread, where `EM_ASM` can't reach the page's `midiOutputDevice`. Realtime bytes are queued in a small ring and flushed from `main_loop__em()` — the same main-thread split as `amy_sequencer_js_hook`.
- **Godot:** `amy_midi.c` is excluded there, so the new functions get no-op stubs in both `godot/src/amy_platform_stubs.c` and the embedded copy in `setup_godot.sh`.
- Docs updated in `docs/midi.md`.

## Verification

- Offline C harness capturing `midi_out()` bytes over rendered audio: `0xF8` arrives at 24 PPQ (43/sec at 108 bpm, measured over 2s running + 1s stopped + 1s restarted), exactly one `0xFA`/`0xFC` per transport edge, nothing but realtime bytes in the stream, nothing at all sent in modes 0/1, and follow mode still advances two sequencer ticks per injected `0xF8`.
- `make test`: all 109 WAV tests pass bit-exact.
- `emcc` compile of `amy_midi.c` and `libminiaudio-audio.c` clean; Godot stub file compiles standalone.

Timing note: clock is emitted from the block-rate sequencer check, so F8 jitter is bounded by one render block (~5.8 ms at 44.1k/256) — the same quantization the sequencer itself has. On web it's additionally quantized to the main-loop frame.

### DIFF
--- a/docs/midi.md
+++ b/docs/midi.md
@@ -43,15 +43,19 @@ For example, to send the wire message `v0f440l1` (sine wave, 440Hz, note on) to 
 
 ## Others
 
-AMY supports MIDI realtime transport for the internal sequencer:
+AMY supports MIDI realtime transport for the internal sequencer, in both directions:
 
 - `0xF8` (Timing Clock): advances the sequencer from external clock input.
 - `0xFA` (Start): starts sequencer playback from tick 0 in external clock mode.
 - `0xFC` (Stop): stops sequencer playback.
 
-External realtime sync is **off by default** — AMY ignores incoming clock/start/stop and keeps its own tempo. Enable it with the C API `amy_external_midi_sync(1)`, or over the wire with `amy.send(external_midi_sync=1)` (wire command `zC1`). Set it back to 0 to return to the internal clock.
+External realtime sync is **off by default** — AMY ignores incoming clock/start/stop and keeps its own tempo. The sync mode is set with the C API `amy_external_midi_sync(mode)`, or over the wire with `amy.send(external_midi_sync=mode)` (wire command `zC`):
 
-MIDI Timing Clock is 24 PPQ. AMY's sequencer runs at 48 PPQ, so AMY advances two sequencer ticks per one MIDI clock pulse.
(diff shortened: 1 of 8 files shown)


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1134

PR #870, opened 2026-07-10
Failsafe signed updated

dpwe:
Verified that, when running on AMYboard, the board does not "failsafe" until the engine is in fact overloaded.  (I modified LoadTestChord.ino to try to do more than 6 notes and it died with "sad tone", but it doesn't die with 6 notes)


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1135

PR #874, opened 2026-07-11
Add amy_external_bus_dsp_hook: custom DSP at the end of each bus' FX chain

bwhitman:
Adds one optional host hook to `amy_config_t`, following the existing `amy_external_*_hook` pattern:

```c
void (*amy_external_bus_dsp_hook)(uint8_t bus, SAMPLE *buf, uint16_t len);
```

Called from `amy_fill_buffer()` at the end of each bus' effects chain (after EQ/chorus/echo/reverb), before buses are mixed to the output. `buf` is `AMY_NCHANS` sequential (non-interleaved) channel blocks of `len` SAMPLEs (S8.23), modified in place. NULL by default: zero behavior change when unset.

Motivation: lets hosts insert user-defined effects into a bus at runtime. This is the AMY half of a Tulip/AMYboard feature where users write a small C effect (e.g. a bit-crusher) as a Python string, compiled on-device (xcc700 + esp-idf elf_loader on ESP32-S3) and dispatched from this hook — design doc in shorepine/tulipcc `docs/user_c_dsp_design.md` (branch `claude/amyboard-c-audio-effects-d1c695`). The custom-oscillator half needs no AMY change (`amy_external_render_hook` already covers it).

`make test`: 109 tests pass, bit-exact (err −100 dB).


dpwe:
Can that really cost 37 microseconds?  Scary.

docs/api.md
```diff
@@ -174,6 +174,7 @@ Hook fields in `amy_config_t`:
 | Hook | Signature | Used by | Description |
 | ---- | --------- | ------- | ----------- |
 | `amy_external_render_hook` | `uint8_t (uint16_t osc, SAMPLE *buf, uint16_t len)` | — | Custom oscillator renderer for redirecting the output waveforms on an osc-by-osc level. Return 1 if handled, in which case that osc does not contribute to the normal output. |
+| `amy_external_bus_dsp_hook` | `void (uint8_t bus, SAMPLE *buf, uint16_t len)` | — | Custom effect processing, called at the end of each bus' effects chain (after EQ/chorus/echo/reverb), before buses are mixed to the output. `buf` is `AMY_NCHANS` sequential (non-interleaved) channel blocks of `len` SAMPLEs each; modify it in place. |
```
dpwe: api.md does not define `SAMPLE` (S8.23 fixed point in an int32_t) nor `output_sample_type` (int16_t).  Also, someone might be surprised that the external bus hook will only be called for buses up to the highest one that has been activated (which may change dynamically).

dpwe: Calling it a "dsp hook" is a bit casual.  I would call it a "postprocess hook".

dpwe (approved):
Comments on naming/documentation.

bwhitman:
[CLAUDE]: Applied in 8d0f470: renamed to `amy_external_bus_postprocess_hook`, api.md now defines `output_sample_type` (int16_t) and `SAMPLE` (S8.23 in an int32_t) at first use, and both the header comment and the hooks table document that the hook fires for buses 0..highest-activated-bus (which grows dynamically). 109 tests still pass bit-exact.

On the +37 µs: with the hook NULL (as in the bench), the added work per block is one load+branch per active bus — single-digit nanoseconds, so the delta can't be the check itself. A uniform +31–49 offset across 1–6 notes smells like a code-layout/alignment shift in the hot `amy_fill_buffer` path from inserting the call site, rather than noise (which wouldn't be so consistent). This push re-runs the bench, which gives us a second sample — if a similar offset shows up again with different code placement, that supports the alignment theory over a real cost.

dpwe:
LGTM

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1146

PR #875, opened 2026-07-11
Avoid transcendentals

dpwe:
Following suggestions from rt_rtos, we remove the calls to exp2f/log2f in freq<>logfreq used in osc initilization, replacing with existing fixed-point lookup tables.  Also added new fixed-point lookup replacements for sin and cos, and use them when recalculating filter coefficients.   Originally I was reusing the sine lookup from osc, but it wasn't fine enough, so now there's a separate 257-point quarter-sine lookup table in log2_exp2_fxpt_lutable.c.

Overall render_us (6x Juno1 voices) down from 4245 to 4112 (-3%).


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1141

PR #881, opened 2026-07-12
Soft-float libcalls are afraid of f´s (sometimes). Added some in the render path to keep them away. 

rt-rtos:
# Suffix float literals to keep FP64 soft-float out of the render path
Follow up to #877. Surfaced during that pass. But required proper verification.

# What this changes

Adds `f` suffixes to float literals whose expressions were being evaluated in
`double`, plus one `(float)` cast on `M_PI`. No functional changes intended -
all 109 reference tests produce byte-identical error figures before and after.

Keeps FP64 libcall out of every `osc`

## Why

An unsuffixed literal like `2.0` promotes its expression to `double`. GCC can
usually fold the constant back to float, but not when the expression ends in
an integer cast - then the arithmetic stays in FP64, which on ESP32-class
targets means soft-float library calls.

The case that matters is in the fixed-point phasor conversions
(`amy_fixedpoint.h`):

```c
#define F2P(f) ((PHASOR)((f) * 2.0 * (float)(1 << (P_FRAC_BITS - 1))))
```

`PHASOR` is an integer, so the multiply chain runs in double:
`__extendsfdf2` + `__adddf3` + `__muldf3` + `__fixdfsi` on every use. Since
`AMY_USE_FIXEDPOINT` is the default build, this reached `render_sine`,
`render_fm_sine`, `render_lpf_lut`, `render_pcm`, `play_delta` and most of the
`compute_mod_*` family. The comment above the macro shows float was the intent
all along - the `2.0` is only there to avoid `1 << 31` overflowing an `int`.

The other edits are the same pattern in smaller doses:

| file | literals |
| --- | --- |
| `filters.c` | biquad clamps `0.0001`, `0.45`, `0.51`, `0.99`; `(float)M_PI` |
| `amy.h` | `ZERO_LOGFREQ_IN_HZ`, `ZERO_HZ_LOG_VAL`, `MIN_FILTER_LOGFREQ`, `CLIP_D` |
| `envelope.c` | `12.375`, `MIN_LEVEL_S`, `ATTACK_RANGE_S`, `BREAKPOINT_EPS`, `EXP_RATE_VAL` |
| `amy.c` | `MAX_DELTA_FILTER_LOGFREQ_DOWN`, the `0.99` filter-coef clamp |
| `oscillators.c` | `0.9`, `0.999` |
| `parse.c` | `1000.0` in `ms_to_samples` |
| `interp_partials.c` | `0.001` dB offset |

On `M_PI`: libc defines it as a `double`, so the `#ifndef M_PI` float fallback
in `filters.c` never fires and `0.01f / (2 * M_PI)` was evaluated in FP64.
That's the LPF residue noted in #877.

## Measured effect

Counting functions that call FP64 soft-float helpers, whole tree, built with
`xtensa-esp32s3-elf-gcc -O2 -DAMY_USE_FIXEDPOINT` (gcc 15.2.0):

**42 before, 10 after.**

The 10 remaining are all `__extendsfdf2` alone next to `printf`/`snprintf` -
the standard varargs float-to-double promotion, not arithmetic. No FP64
arithmetic remains in the audio path.

## Correctness

All 109 reference tests produce byte-identical error figures against an
unmodified build.

## Why the Makefile is included

The `src/patches.h` rule scrapes `#define`s from `src/amy.h` into
`amy/constants.py`, and its sed only strips a trailing `f` at end of line. A
suffixed define followed by a comment
(`#define ZERO_LOGFREQ_IN_HZ 440.0f  // 261.63`) would leak `440.0f` into
Python and break the build, so the sed now also handles a suffix followed by
whitespace.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ HEADERS_BUILD := $(filter-out src/patches.h,$(HEADERS))
 PYTHONS = $(wildcard *.py)
 
 src/patches.h: $(PYTHONS) $(HEADERS_BUILD)
-	cat src/amy.h  | sed -e 's@^//.*@@' | egrep 'define +[^ ]+ +[.0-9-]+' | sed -e 's/\([.0-9]\)f$$/\1/' | awk '{print $$2 "=" $$3}' > amy/constants.py
+	cat src/amy.h  | sed -e 's@^//.*@@' | egrep 'define +[^ ]+ +[.0-9-]+' | sed -e 's/\([.0-9]\)f\([ 	]\|$$\)/\1\2/' | awk '{print $$2 "=" $$3}' > amy/constants.py
 	${PYTHON} -m amy.headers
 
 %.o: %.c $(HEADERS) src/patches.h
--- a/src/amy.c
+++ b/src/amy.c
@@ -263,7 +263,7 @@ void config_echo(uint8_t bus, float level, float delay_ms, float max_delay_ms, f
     amy_global.bus[bus]->echo.level = F2S(level);
     amy_global.bus[bus]->echo.delay_samples = delay_samples;
     // Filter is IIR [1, filter_coef] normalized for filter_coef > 0 (LPF), or FIR [1, filter_coef] normalized for filter_coef < 0 (HPF).
-    if (filter_coef > 0.99)  filter_coef = 0.99;  // Avoid unstable filters.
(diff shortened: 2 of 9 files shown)


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1143

PR #888, opened 2026-07-13
sprint_event: avoid %hu format (unsupported by MicroPython internal printf)

bwhitman:
## Problem

`sprint_event` in `src/patches.c` printed the `osc` field with `PRIu16`, which expands to `%hu`. MicroPython's internal printf (`mp_vprintf`) has no `h` length-modifier support, so in any MicroPython port where `MICROPY_USE_INTERNAL_PRINTF=1` overrides `snprintf` (the default; `micropython/shared/libc/printf.c` routes it into `mp_vprintf`), formatting reaches `assert(*fmt == '%' || !"unsupported fmt char")` in `py/mpprint.c` and aborts on debug builds — triggered by anything that serializes events: `tulip.amy_dump_state()`, `amyboard.update_sketch_knobs()`, etc.

Affected in practice: Tulip Linux desktop builds and the AMYboard VCV Rack plugin build. Tulip **macOS** desktop happens to be immune: Darwin's fortify headers rewrite `snprintf` calls to `___snprintf_chk` (system libc) regardless of `-U _FORTIFY_SOURCE`, bypassing the MicroPython override — verified live at the desktop REPL.

## Fix

Print the osc field with `%u` and an `(unsigned)` cast (varargs promote `uint16_t` to `int` anyway, so this is value-identical). These were the only two `%hu` sites reachable through `snprintf` in AMY; all other formats used (`%d`/`%u` via 32-bit PRI macros, `%.3f`, `%s`) are supported by `mp_vprintf`.

## Testing

`make test`: 109 tests pass.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1245

PR #892, opened 2026-07-15
AMY_HOST_MIDI define + Windows miniaudio fixes

bwhitman:
## AMY_HOST_MIDI

A properly-named define for "the embedding host owns the MIDI device layer" (`run_midi`/`stop_midi`/`midi_out`), replacing the `-DMACOS`-on-every-platform trick the AMYboard VCV Rack plugin was using. `MACOS` implies `AMY_HOST_MIDI` (macos_midi.m is that host layer), so mac desktop builds are unchanged; `amy_midi.c`'s device layer is now gated on `AMY_HOST_MIDI` instead of `MACOS`.

## Windows miniaudio fixes

Found bringing Tulip Desktop up on Windows (shorepine/tulipcc#1155):

- Windows device periods aren't a multiple of `AMY_BLOCK_SIZE` (measured 882 frames on WASAPI shared, 1323 on DirectSound), so the output ring's stock one-block lead oscillates down to zero → underrun crackle. The lead is now sized to the negotiated period (+margin), and set **before** `ma_device_start` (the ring is also zeroed before start, fixing a startup-crackle race).
- `OUTPUT_RING_FRAMES` grown 8→16 blocks to hold the larger lead.
- Windows prefers DirectSound over shared-mode WASAPI (measurably fewer glitches in VM testing; instrumentation showed the stream itself clean on both — no ring starvation, callbacks 0.2ms against a 20ms budget).

mac/linux behavior is unchanged (they force period == `AMY_BLOCK_SIZE`, where the one-block lead is exact).

`make test`: 109 tests pass. Validated in the tulipcc Windows desktop build and all three VCV Rack platform builds.

### DIFF
--- a/src/amy_midi.c
+++ b/src/amy_midi.c
@@ -467,7 +467,7 @@ void midi_out(uint8_t * bytes, uint16_t len) {
 }
 #endif
 
-#if !defined(MACOS) && !defined(__EMSCRIPTEN__) // this code is for NOT macos desktop , which is in macos_midi.m
+#if !defined(AMY_HOST_MIDI) && !defined(__EMSCRIPTEN__) // device layer: skipped when the host owns MIDI (AMY_HOST_MIDI — macos desktop via macos_midi.m, VCV plugin, etc; see amy_midi.h)
 
 // "run_midi" sets up MIDI on MCU platforms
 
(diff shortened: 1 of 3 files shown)


bwhitman:
This is a no-op on anything but windows , will merge

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1157

Issue #889, opened 2026-07-14
ESP32-S3: PIE (SIMD) block clear/copy in the render path - PR wanted?


rt-rtos:
#### Setup:
AMY (fixed-point, v1.2.31) on ESP32-S3 

I spent some time investigating what the S3's PIE unit (128-bit SIMD) can do for the render path.
The result is small but real, gracefully does nothing on every other
target and avoids pulling in esp-dsp as a dependency.
 Probing interest for a PR before I shape one.

### What I found

Almost nothing in AMY is PIE-eligible, and I think that negative result is
worth recording too:

- PIE multiplies only on 8/16-bit lanes (`EE.VMULAS.S16` into 40-bit
  QACC/ACCX). 32-bit lanes get add/sub/shift/min/max/compare only, no gather,
  no float SIMD. `SAMPLE` is s8.23 = int32, so every hot-path multiply
  (`MUL8_SS`, `SMULR6`, `top16SMUL`) is a 32x32 product PIE cannot vectorize.
- The LUT oscillators are gather-indexed (wavetable walked by phase
  accumulator) and the biquads/EQ/echo/chorus/reverb are recurrence-bound
  (IIR feedback), so neither vectorizes on any SIMD unit. esp-dsp reaches the
  same conclusion: its own ESP32-S3 biquad (`dsps_biquad_f32_aes3.S`)
  contains zero PIE instructions.
- I also tried vectorizing `scan_max()` / `block_norm()` (multiply-free
  reductions, so they looked eligible). On-target A/B said no: flat on most
  scenes, up to ~1% slower on filter-heavy ones. Nearly every call is on a
  tiny buffer (`scan_max(w, 4)`, the 8-entry `filter_delay`) where vector
  setup costs more than the scalar loop.

What is left is the bulk block clears and copies, and those do pay:

| Call site | Change |
|---|---|
| `amy.c` `amy_render()`: `fbl` and `per_osc_fb` clears | `bzero` -> 128-bit `EE.VST.128` block fill |
| `amy.c`: `chorus.delay_mod` clear | same |
| `algorithms.c` `zero()` / `copy()` (FM operator scratch) | block fill / `EE.VLD.128`+`EE.VST.128` block move |

The FM scratch is where the win is. 

#### Conditions:

- ESP32-S3 @ 240 MHz 
- ESP-IDF
- fixed-point build
- 48 kHz
- 256-sample blocks; 
- metric is median CPU
[...]

PR #893, opened 2026-07-15
 Inline PIE asm in algorithms.c zero()/copy() [updated]

rt-rtos:
### Update a96d574 following review

Hope this alligns with expressed intent.

- Scratch allocation flattened (suggested by @dpwe): `scratch` is now one
  contiguous 16-byte-aligned allocation of `AMY_CORES * 3` sample blocks
  instead of a `SAMPLE***` pointer-array-of-pointer-arrays. One malloc instead
  of `1 + AMY_CORES * 4`, `algo_deinit()` is a single `free()`, and
  `render_algo()` derives `BUS_ONE`/`BUS_TWO`/`SCRATCH` with constant offsets
  instead of two dependent pointer loads. Each interior block stays 16-byte
  aligned because `AMY_BLOCK_SIZE * sizeof(SAMPLE)` is a multiple of 16, so
  the PIE vector path is unaffected.

- Allocation count: on dual-core, algo scratch drops from 9 heap allocations
  to 1. Fewer per-block heap headers, and aligned allocation over-allocates
  per call, so one aligned block means less padding.
- Fragmentation: S3 is notorious for heap fragmentation.
  One long-lived block claimed at init instead of nine small
  ones interleaved with whatever else allocates at startup - friendlier to
  heap.

*Note about IDF allocator quirk*
> **The "Largest Heap First" Allocator Flaw**
>A known quirk in the ESP-IDF capability allocator exacerbates this issue. By default, standard malloc calls aggressively target the largest available free block first. If your application makes dozens of small allocations (like a single string or JSON token), they are placed directly into your primary 200 KB pool instead of filling up the 20 KB or 32 KB pools first. This rapidly shreds the only block capable of handling large buffers.

- Lifecycle: `algo_deinit()` is a single `free()` - no partial-teardown
  states..
- Shape: `SAMPLE*` with constant offsets instead of `SAMPLE***` - the
  layout (`BUS_ONE | BUS_TWO | SCRATCH` per core, contiguous) is visible in
  the code, and the single base alignment guarantee covers every block.

---

### Original

Implements the ESP32-S3 PIE block clear/copy from #889, in the inline shape we
agreed on there: inline asm inside `algorithms.c`'s `zero()`/`copy()`, gated on
`__XTENSA__ && CONFIG_IDF_TARGET_ESP32S3`, falling back to `bzero`/`bcopy`
everywhere else. No new source files, no `.S`, no vendored assembly.

**What**
- `zero()`/`copy()`: 128-bit `ee.vst.128` / `ee.vld.128` loop for the FM
  operator scratch on the S3; libc fallback for unaligned or non-S3.
- `malloc_caps_block()` (in `amy.c`): 16-byte-aligned allocator so the scratch
  takes the vector path. Aligned-only by design - unaligned just uses libc.

**Provenance**: the aligned inner loop follows esp-dsp's
`dsps_memset_aes3.S` / `dsps_memcpy_aes3.S` approach, but is original code (no
lines copied) and omits their scalar head/tail handling entirely - so it needs
no `.S` file no extra apache license.

*(Though I would love if someone more experienced with licenses
could take a look to be sure - the comment above the inline asm
lays out the specifics)*

---

240 MHz · fixed · 48000 Hz / block 256 · free pacing · 5 captures/side · 3 passes · block budget 5333 us

| Scene | A cyc | B cyc | delta | noise | headroom | verdict |
|---|---:|---:|---:|---:|---|---|
| juno6 | 1,159,437 | 1,157,434 | -0.17% | ±0.01% | 9.4% -> 9.6% | small |
| dx76 | 765,569 | 688,524 | **-10.06%** | ±0.01% | 40.2% -> 46.2% | IMPROVEMENT |
| saw_lpf6 | 442,455 | 442,322 | -0.03% | ±0.01% | 65.4% -> 65.5% | small |
| fx_sine8 | 386,551 | 386,941 | +0.10% | ±0.00% | 69.8% -> 69.8% | small |
| sine8 | 162,846 | 163,735 | **+0.55%** | ±0.01% | 87.3% -> 87.2% | REGRESSION(Layout) | 
| pcm4 | 141,812 | 141,886 | +0.05% | ±0.02% | 88.9% -> 88.9% | small |
| idle | 34,967 | 34,964 | -0.01% | ±0.00% | 97.3% -> 97.3% | small |

**7 scenes: 1 improved, 1 regressed(layout variance), 5 inconclusive** (within noise, below threshold, or unjudgeable).

### DIFF
--- a/src/algorithms.c
+++ b/src/algorithms.c
@@ -74,13 +74,51 @@ const struct FmAlgorithm algorithms[33] = {
 
 // a = 0
 static inline void zero(SAMPLE* a) {
-    bzero((void *)a, AMY_BLOCK_SIZE * sizeof(SAMPLE));
+    const size_t nbytes = AMY_BLOCK_SIZE * sizeof(SAMPLE);
+#if defined(__XTENSA__) && defined(CONFIG_IDF_TARGET_ESP32S3)
+    // ESP32-S3 PIE: clear the FM-operator scratch block with a 128-bit vector
+    // store loop. This block clear/copy is the render path's one real PIE win
+    // (~10% on a 6-op FM voice); nothing else in AMY vectorizes (int32 SAMPLE has
+    // no 32-bit vector multiply, oscillators gather, filters recur). The scratch
+    // is 16-byte aligned (malloc_caps_block in algo_init) and nbytes is a compile
+    // -time multiple of 16, so the guard passes; anything unaligned falls to bzero.
+    if ((((uintptr_t)a | nbytes) & 15u) == 0) {
+        void *pp = a;  // ee.vst.128.ip post-increments the address register
+        __asm__ volatile(
+            "ee.xorq q0, q0, q0\n\t"
+            "loopnez %1, .Lamy_zero%=\n\t"
+            "ee.vst.128.ip q0, %0, 16\n"
+            ".Lamy_zero%=:"
+            : "+&r"(pp)
+            : "r"(nbytes / 16)
+            : "memory");
+        return;
+    }
+#endif
+    bzero((void *)a, nbytes);
 }
 
 
-// b = a 
+// b = a
 static inline void copy(SAMPLE* a, SAMPLE* b) {
-    bcopy((void *)a, (void *)b, AMY_BLOCK_SIZE * sizeof(SAMPLE));
+    const size_t nbytes = AMY_BLOCK_SIZE * sizeof(SAMPLE);
+#if defined(__XTENSA__) && defined(CONFIG_IDF_TARGET_ESP32S3)
+    // ESP32-S3 PIE: copy the block with paired 128-bit vector load/store (see zero()).
+    if ((((uintptr_t)a | (uintptr_t)b | nbytes) & 15u) == 0) {
+        const void *s = a;
+        void *d = b;
+        __asm__ volatile(
+            "loopnez %2, .Lamy_copy%=\n\t"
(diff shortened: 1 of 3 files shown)


dpwe:
This looks great. 

I added an effort to reduce the total number of mallocs (and remove the double-indexing into scratch) which seems to have helped my performance numbers.  See https://github.com/shorepine/amy/tree/pie-simd-inline

rt-rtos:
Awesome, glad it helped. Excited to see your refinements. Currently resolves to my two commits - maybe not pushed yet?

dpwe:
I was thinking you could add the changes to scratch[] to your PR, then I would commit that.

I am sensitive to attribution, maybe I'm overdoing it.

rt-rtos:
> I was thinking you could add the changes to scratch[] to your PR, then I would commit that.
> 
> I am sensitive to attribution, maybe I'm overdoing it.

It´s appreciated.

PR body has an edit with an update.

dpwe:
Maybe you thought I was being cryptic or something.   No, I just forgot to commit my change which avoided the multiple mallocs.  But yours is neater (I was still allocating scratch[NUM_CORES * 3]) so this is great.  Thanks again!

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1161

rt-rtos:
Figured as much. Honestly didn't mind. It was a useful excercise in working from somewhat amgibuous instructions.

PR #896, opened 2026-07-15
Add amy_external_midi_output_hook to catch AMY's MIDI out

### LINKED ISSUE #895: Please implement a way to catch Amy's midi out
I use my custom ESP32S3 with Amy and I managed to create a BT MIDI in and out. Usually I use my ESP through UART MIDI out on a pin, but with a call [extern "C" void amy_event_midi_message_received(uint8_t *data, uint32_t len, uint8_t sysex, uint32_t time);] I pass BT MIDI in data to Amy. That works.

Now I need a way to catch Amy's MIDI out data (even if no serial MIDI out pin defined in config), to pass data on BT MIDI out. In that way I assume that I could connect through BT (and UART at the same time) and for eg. try things or create new patch with the online Amy editor (or my custom way with CC controls) without the need to disconnect it from my main synth.

I think it would be as simple as config the amy_external_midi_input_hook. For eg. an amy_external_midi_output_hook which will be called inside midi_out(...) method to pass all data to my function.

bwhitman:
Fixes #895

Adds `amy_config.amy_external_midi_output_hook`, mirroring the existing `amy_external_midi_input_hook`, so hosts can catch everything AMY sends over MIDI out and forward it to transports AMY doesn't drive itself — e.g. BLE MIDI on a custom ESP32-S3 board, as requested in the issue.

```c
void my_midi_out_hook(uint8_t *bytes, uint16_t len) {
    // e.g. forward to BT MIDI out
}

amy_config_t amy_config = amy_default_config();
amy_config.amy_external_midi_output_hook = my_midi_out_hook;
amy_start(amy_config);
```

## How it works

- New `midi_out_external_hook()` helper in `amy_midi.c` (platform-neutral section) calls the config hook if set.
- Every in-repo `midi_out()` implementation calls it first, **before — and regardless of — any device interface configured in `config.midi`**: the MCU/device layer (UART / USB gadget), the emscripten/WebMIDI build, and `macos_midi.m`. So the hook fires even with no serial MIDI out pin configured, per the issue.
- The hook sees all of AMY's MIDI output: note on/off forwarding, realtime clock (0xF8/0xFA/0xFC), and sysex responses (`zD` dumps, transfer frames).
- Hosts that own the whole MIDI layer (`AMY_HOST_MIDI`, e.g. the VCV plugin) define their own `midi_out` and can call the helper themselves; nothing changes for them by default.
- Defaults to `NULL` in `amy_default_config()`; documented in `docs/api.md`. Godot builds are unaffected (`amy_midi.c`/`macos_midi.m` are excluded there and nothing in the stubs references the helper).

Verified with a clean macOS desktop build (`make`), which compiles the neutral `amy_midi.c` section, `macos_midi.m`, and the new config field together.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -183,6 +183,7 @@ Hook fields in `amy_config_t`:
 | `amy_external_coef_hook` | `float (uint16_t channel)` | — | Provide external coefficient values (e.g. CV input). |
 | `amy_external_block_done_hook` | `void (void)` | — | Called after each audio block is rendered. |
 | `amy_external_midi_input_hook` | `void (uint8_t *bytes, uint16_t len, uint8_t is_sysex)` | — | Called when MIDI bytes are received. |
+| `amy_external_midi_output_hook` | `void (uint8_t *bytes, uint16_t len)` | — | Called with every run of bytes AMY sends over MIDI out (notes, clock, sysex responses), before — and regardless of — any device interface configured in `midi`. Use it to forward AMY's MIDI output to a transport AMY doesn't drive itself, e.g. BLE MIDI. May be called from the render/sequencer task; keep it fast. |
 | `amy_external_sequencer_hook` | `void (uint32_t tick_count)` | — | Called on each sequencer tick. |
 | `amy_external_fopen_hook` | `uint32_t (char *filename, const char *mode)` | `zT`, `zD`, `zF` | Open a file on host disk. Returns opaque handle. |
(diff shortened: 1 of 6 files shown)


bwhitman:
dan said ok

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1278

PR #902, opened 2026-07-16
Clear midi note cmds

dpwe:
There was a bug where even though we *meant* to clear all the midi mapping commands (CCs and notes) when resetting a synth (with `num_voices=0`), the synth number had already been cleared before we passed it to the reset routine, so they were not.

In the process, I found that there was scheduling skew between when oscs were being cleared and when they were being set up again, during synth patch changes.  I think it's all aligned now.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1162

PR #904, opened 2026-07-16
PCM: respect trigger_phase (P) as the sample start point at note-on

bwhitman:
`pcm_note_on()` unconditionally reset phase to 0, clobbering a `trigger_phase` that arrived in the same event as the note-on (`play_delta` applies trigger_phase to phase *before* `osc_note_on` runs). The docs already said `P` "also works on the PCM buffer", but at note-on time it didn't.

## What changed

- A set `trigger_phase` now becomes the PCM start phase at note-on (`start_frame / 2^PCM_INDEX_BITS`) and is **consumed** by that note-on, so it can't leak into a later note that reuses the osc — e.g. a drum-synth voice being reallocated to a different drum whose note-on template doesn't carry a `P` of its own. Post-note-on `P` messages (the gamma9001 pattern) behave as before: the phase delta still applies immediately.
- `docs/api.md`: documented the PCM note-on semantics of `P`.
- New `TestPcmTriggerPhase` (+ reference): a `P` in the note-on event starts the sample partway in; the next note-on without a `P` starts from 0 again.

## Why

This lets a MIDI note mapping (`io...`) bake a per-drum sample start offset into its note-on template. The AMYboard editor's new drum kit knob view (tulipcc PR to follow) uses this for its per-drum Offset knob, like the gamma9001 demo's Offset — verified end-to-end in the web simulator (kick at offset 0.9 renders near-silence vs full attack at 0).

`make test`: 112 tests pass.

## Update: kit note templates carry an explicit filter-off (G0)

Second commit: `make_drums_patch()` now bakes `filter_type=0` (`G0`) into every kit note template. Kit voices are reused across drums and osc params persist on the osc, so a template that doesn't mention `filter_type` inherits whatever the last drum on that voice set — a per-drum editor putting an LPF on one drum would smear it onto every other drum in the kit. Deliberately **not** pan: templates layer over the caller's event, so a baked `Q0.5` would clobber per-note pan (`amy.send(synth=10, note=50, pan=...)` — caught by `TestSynthDrumsPanning` when I tried). 112 tests pass.


bwhitman:
Dans says OK

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1167

PR #910, opened 2026-07-16
Drum kits: per-drum gain in the note map's velocity scale, freeing amp CONST as the channel level

bwhitman:
Fixes the channel Level control having no effect on drum kit channels (reported against the AMYboard Online channel strip, but it affects any `i<ch>v0a<val>` — MIDI CC 7 included).

## The bug
The kit patches (K258, K384+) baked each drum's level-balancing gain into the note template's amp CONST (`n60l%va4.02...`), so every hit rewrote the osc's amp coefficient — a per-channel level set on the synth was clobbered by the next drum.

## The fix
Move the gain into the note mapping's **linear velocity scale** (`io<note>,0,0,<gain>,0,...`). Amp coefficients combine in the log domain (`amp_combine_controls`), so `level · (gain·vel) == (level·gain) · vel` — loudness-identical at every velocity. The previously-considered alternative of moving the gain to the **vel coefficient** (`a,,4.229`) would be wrong: that coefficient multiplies the log-compressed velocity, i.e. it's an *exponent* on velocity (`vel^4.229`), not a gain.

The kit header's `amp=5` is dropped as well, so amp CONST stays at its reset default 1.0 and behaves exactly like the melodic patches' channel level. It was inert for the balanced kits (every hit overwrote it) but load-bearing for the pcm_tiny K258 variant, which keeps its ×5 via the velocity scale instead.

## Tests
- `make test`: 113 pass.
- The three existing drum test refs were regenerated: the audio differs from the old refs only by LUT rounding (err ≈ −63 dB relative to signal; hits at vel=1 are bit-identical).
- New `TestSynthDrumsLevel` locks the fix in: the same snare at channel levels 0.2 / 1.0 / 1.8 with two hits per level. Measured peak scaling is exactly 5.00× for the 0.2→1.0 step, and the level survives across note-ons (that's the regression).

Companion tulipcc PR updates the AMYboard Online drum editor (per-drum vol knob becomes a multiplier on the baked gain, with migration of previously saved sketch lines).


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1169

PR #913, opened 2026-07-17
Drums synth changed to have one voice with 38 voice-specific oscs.

dpwe:
This patch significantly changes how AMY drums patches (258 and above) work.  Whereas before they had typically 6 voices of one osc each which were re-used for the next note whatever it was, now they have one voice with (e.g.) 38 oscs, one for each distinct drum sound (the "voice" is just acting as a container for all the individual, unique drum oscs).  Each drum osc is initialized with the appropriate PCM preset, and then `midi_note_cmd`s are set up to map the GM drum notes to send note-ons (with the matching base_note) to the appropriate osc within the single voice.  

Because of this "container" and dedicated-osc structure, drum-patch synths should **only have one voice** (`num_voices=1`).  Because the voice uses a lot of oscs (e.g. 38), it's easy to run out of resources if you accidentally try to allocate, for instance, 6 voices.  Unfortunately, this was common with drum patches in the past.  **All such drum-patch setups with `num_voices` greater than 1 must now be fixed to work with this pull request.**

Drum note-on events should work as before, but now you can also persistently change the parameters (including pan, volume, filter effects, ...) of a particular drum by directly addressing the relevant osc.  For instance, looking at patch 384 (Gamma9001 TR-808), the hand clap is GM note 39.  To permanently set the pan of this handclap to 0.1, we would `amy.send(synth=10, note=39, pan=0.1)`, assuming that synth 10 had been set up as a drum patch with e.g. `amy.set(synth=10, num_voices=1, patch=384)`.

The semantics of sending a (non-voice-on) command to a synth **when no osc is specified** has changed.  Before, if no osc was specified, e.g. `amy.send(synth=1, amp=0.5)`, the command was sent only to osc 0 in each voice.  Now omitting the osc means to send the command to *all* oscs in the voice; to preserve the previous behavior you should *always* specify osc 0 when directing commands to the control (`SILENT`) osc in Juno and AMYboard Web patches (e.g. to control overall level, ADSR, and filter) like `amy.send(synth=1, osc=0, amp=0.5)`.  However, this change means you can scale the level of *all* the oscs in a drum synth with `amy.send(synth=10, amp=0.5)` since this modification is broadcast to all the per-drum oscs.

Happily, it seems that the AMYboard Web UI already explicitly specifies osc 0 when changing the volume and other control osc parameters.  For drum channels only, we will need to *remove* the reference to osc 0.

The implementation of the AMYboard web drum editor will have to be changed significantly.  Currently, the per-drum parameter knobs (pitch, offset, pan) rewrite the `midi_node_cmd` (`io`) for that particular note to include the modified parameters.  Now, instead, these knobs should directly set the matching parameter on the corresponding osc, referenced by the MIDI note number, e.g. `i10n39p0.1` (`synth=10, note=39, pan=0.1`).  Then, the `io` commands remain fixed (e.g. `i10io39,0,0,1,0,i%in60v%v`) and do not to be rewritten.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1173

PR #918, opened 2026-07-17
Add per-instrument level (synth_level / iV), applied to osc audio at render

bwhitman:
Adds a per-synth overall level, as discussed with @bwhitman for the AMYboard channel Level sliders (follow-on to #913): `amy.send(synth=10, synth_level=0.5)`, wire `i10iV0.5`. Float, default 1.0 when unset, stored on the instrument.

**Where it applies:** at render, to every osc of the synth — but at the **audio output stage** (the rendered per-osc buffer, just before the pan/bus mix), not on each osc's amp. Scaling every osc's *amp* squares the level on layered patches (a Juno control osc's amp already multiplies into its carriers — measured -24 dB for level 0.25 with that approach; the output-stage scale gives the exact -12 dB).

**Implementation**
- `instrument_info.level` + `instrument_get/set_level()`; `iV` parsed as a synth-layer command (`V` was free) and applied at receive time like the other `i` commands (`if`, `im`, `ip`)
- A per-(amy)voice mirror `voice_level[]` makes the render lookup O(1) (`osc_to_voice[osc] → voice_level[voice]`); refreshed on instrument create/free/level change, so unowned voices/oscs stay at 1.0
- `sprint_event` prints it, and `yield_synth_events`' setup preamble includes `iV<val>` when ≠ 1, so synth readback (web editor state sync) carries it
- Python API gains `synth_level` (`iVF`); Godot map regenerated via `make godot-api`
- The K257 amyboard patch's baked CC 7 mapping becomes `i%iiV%v` — `TestPreset257MidiCCs` output is **unchanged** (a linear amp const and the render-stage scale are numerically identical there), and CC 7 now also works on drum-kit synths, which have no control osc for the old `v0a` form
- Docs: `synth.md` gets a `synth_level` example (and its stale `num_voices=6` drum example is fixed to 1 per #913)

**Tests:** all 114 pass (113 existing unchanged + new `TestSynthLevel`: same Juno patch at level 1 vs 0.25 on two synths, plus GM drums at 0.5). Level ratios verified exact on a deterministic PCM kick: 0.4999× for `iV0.5`, 2.0002× for `iV2`.

The tulipcc side (all Level sliders → `iV`, drum vol knobs as bare osc amps, legacy level-line migration) rides on shorepine/tulipcc#1174.


bwhitman:
@dpwe Done in 7ecf09e — the per-sample loop is gone. `mix_with_pan` already ramps per-sample L/R pan gains across the block, so the level now folds into the four gain endpoint floats (L/R × start/end) before `F2S`, and the existing ramp carries it for free: ~4 float multiplies per block, independent of level. (Mono has no gain ramp to ride, so it keeps a per-sample multiply, gated on `level != 1`.)

Side benefit: the old loop scaled with `MUL4_SS` (4 integer bits → would saturate at levels above ~8 in fixed-point builds); the pan path's `MUL8_SS` has the headroom for the full 0–9.9 slider range.

`make test`: 114 pass. TestSynthLevel's ref regenerated — collapsing two quantized multiplies into one shifts the fixed-point rounding by about −90 dB.

bwhitman:
Reverted my mix_with_pan fold in 71f8cf6 — branch is clear for your version. (`make test`: 114 pass, back on the original refs.)

bwhitman:
Reapplied the mix_with_pan fold in 1b7dcdf per discussion — level rides the pan-gain endpoints again (~4 float mults/block in stereo, per-sample only in mono when level != 1, MUL8_SS headroom for the 0-9.9 range). `make test`: 114 pass with the single-quantization TestSynthLevel ref.

bwhitman:
dan says go for it (and we may revisit this later)

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1177

Issue #915, opened 2026-07-17
MIDI on unused channels may still be preprocessed before being discarded


chrgthb:
Hi again, and thanks for AMY!

I suspect incoming MIDI is still being preprocessed on all channels, even if I only created synths on two channels.

I am using custom ESP32S3 AMY as a bytimbral synth with a groovebox (Novation Circuit Tracks) that sends MIDI on multiple channels, and I cannot selectively stop transmission on the channels not meant for AMY. In practice, if all channels are active during playback, the audio starts to stutter. If I disable the extra channels and leave only the two AMY channels active, playback is clean and stable. Because of that, I suspect the issue is not raw ESP performance, but CPU time being spent on MIDI messages for channels that are not actually used by configured synths.

The most suspicious place seems to be [amy_event_midi_message_received()](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) in src/amy_midi.c, where normal MIDI handling appears to happen before the external MIDI hook. [midi_message_handler_to_queue()](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) in src/midi_mappings.c also looks relevant, since MIDI messages seem to be routed there before unused-channel events are effectively dropped.

Would it be possible to discard non-target MIDI channels earlier in the pipeline? If that is not the right default behavior, a user-configurable channel filter or whitelist would be very helpful for setups where external gear always sends unavoidable traffic on additional channels.

Thanks

dpwe:
OK, I've made the processing of MIDI messages quite a bit more heavyweight lately.  I'll work on fast exit and report back.

dpwe:
#924 and later should improve this a lot.  I'd love to hear if it fixes your problems!

chrgthb:
Thanks! I've looked over the changes and they seem to work as intended, but just to be safe (since I'm just a hobby programmer 😃 ):

'amy_external_midi_input_hook' should still be called for unused MIDI channels, just in case someone wants to implement a MIDI-Thru function for AMY.

dpwe:
Thanks for looking!  It looks to me as if that part of `amy_midi_received` is still executed, even for midi packets from channels marked "inactive". 

I prefer tests to trusting my eyes, but sadly we don't have automated tests for the hooks.

dpwe:
Fixed in #924

chrgthb:
[...]

PR #920, opened 2026-07-18
Introduce midi_active_channel*; skip most processing if midi channel not active.

dpwe:
This is in response to #915 which noted severe slow-down from lots of non-AMY-directed MIDI traffic.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1178

PR #924, opened 2026-07-18
midi_mappings.c: Store mappings in 2 x 16 lists for fast checking.

dpwe:
Previously, all the `midi_cc` and `midi_note_cmd` objects for all the channels were stored in a single linked list.  When a MIDI event came in (CC or note), it had to be checked against the whole list to see if there was a chan & type match.

With this change, there are 16 per-channel lists of MIDI CC mapping configs, and another 16 per-channel lists of MIDI note command mappings.  Now, when a midi CC or note-on/off comes in, even for an active midi channel, the search list is shorter, and large sets of mappings for one channel (e.g., the note mappings for the drum sets) don't slow down the processing of MIDI events for other channels.

### DIFF
--- a/src/amy.c
+++ b/src/amy.c
@@ -494,6 +494,7 @@ int8_t global_init(amy_config_t c) {
         ++bus_configs;
     }
     buses_reset();
+    midi_mappings_init();  // Make sure arrays are clear.
     
     amy_init_lock();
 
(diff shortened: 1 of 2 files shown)


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1180

PR #926, opened 2026-07-18
Make AMY state dumps complete and replayable

mateusz28011:
## Summary

- emit volume, EQ, reverb, chorus, and echo state exactly once per **active** bus (`0..amy_global.highest_bus`), independent of instrument enumeration
- expose public `amy.dump_state()`, arbitrated like `get_synth_commands`: `c_amy.dump_state` under CPython, `tulip.amy_dump_state` under MicroPython (Tulip wrapper submitted separately)

## Motivation

`amy_dump_state` currently cannot be used as a complete snapshot and replay interface:

- global bus state depends on which instrument is enumerated first,
- a bus with FX configured but no instrument currently routed to it is not represented,
- there is no public Python entry point for the dump.

Buses above `highest_bus` are deliberately left out of the dump — replaying them would activate them, adding render cost for buses nothing uses.

Instrument bus routing in dumps comes from the synth preamble fixed in #931; this PR is dump-layer only and composes with it (no shared files).

## Verification

- `make test` — 115 upstream AMY tests pass
- `python3 -m unittest -v tests.test_state_dump` — 4/4 pass
- all four buses with distinct volume, EQ, reverb, chorus, and echo state survive dump → reset → replay exactly
- inactive buses stay out of the dump
- `amy.dump_state()` matches `c_amy.dump_state()`
- downstream real-WASM readback and live-edit tests pass

### DIFF
--- a/amy/__init__.py
+++ b/amy/__init__.py
@@ -6,7 +6,13 @@ import time
 def _get_synth_commands_stub(synth, include_fx=False):
     return []
 
+def _dump_state_stub():
+    # Unlike get_synth_commands, an empty result would be indistinguishable from
+    # genuinely empty state, so fail loudly when the build has no dump support.
+    raise RuntimeError("amy.dump_state() is not available in this build")
+
 _get_synth_commands = _get_synth_commands_stub
+_dump_state = _dump_state_stub
 try:
     import c_amy as _amy  # Import the C module
     live = _amy.live
@@ -15,6 +21,7 @@ try:
     _ticks_ms = _amy.ticks_ms
     _render_load = _amy.render_load
     _set_render_load_threshold = _amy.set_render_load_threshold
+    _dump_state = _amy.dump_state
 except (ImportError, AttributeError):
     # C module is not required? not available?
     # I'm guessing this might mean we're on Micropython?
@@ -25,6 +32,8 @@ except (ImportError, AttributeError):
         _ticks_ms = tulip.amy_ticks_ms
         _render_load = tulip.amy_render_load
         _set_render_load_threshold = tulip.amy_set_render_load_threshold
+        # Older tulip builds predate the wrapper; keep the loud stub there.
+        _dump_state = getattr(tulip, "amy_dump_state", _dump_state_stub)
     except (ImportError, AttributeError):
         pass  # Not available (e.g. web build); _get_synth_commands returns []
 
@@ -583,6 +592,13 @@ def get_synth_commands(synth, patch_num=None, dest_synth=None, num_voices=6, inc
         prologue = [prefix + "i%div%din%dZ" % (dest_synth, num_voices, num_oscs)]
         prefix += "i%d" % dest_synth
     return "\n".join(prologue + [prefix + command for command in commands])
+
+"""
+    Reading back the complete engine state
+"""
+def dump_state():
+    # Full replayable engine state as newline-separated wire commands.
+    return _dump_state()
 
 """
     Simulate CV input from an osc (testing support)
--- a/src/amy.h
+++ b/src/amy.h
@@ -1123,6 +1123,7 @@ extern int sprint_event(amy_event *e, char *s, size_t len, bool wirecode);
 extern void fprintf_event_stderr(amy_event *e);
 extern void *yield_synth_events(uint8_t synth, struct amy_event *event, bool include_fx, void *state);
 extern void *yield_synth_commands(uint8_t synth, char *s, size_t len, bool include_fx, void *state);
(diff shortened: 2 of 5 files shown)


dpwe:
I appreciate the new Python "dump_state" command.  Are you planning to make this available in MicroPython and/or WASM as well as CPython?  It's a little involved - you have to add a matching micro python C wrapper in `tulipcc/tulip/shared/modtulip.c`, then put logic in `amy/__init__.py` to arbitrate depending on the environment in which its launched.  Making it work under WASM is even harder (Brian's domain) but maybe that doesn't matter.

Did you experience that instrument routings are not being saved?  From the look of it, calling `yield_synth_state` with `include_fx=true` should emit a bus-related line which I think correctly saves the bus assignment of that instrument.

Agreed that dumping with multiple instruments on the same bus redundantly specifies the parameters for that bus with each instrument.  However, this shouldn't lead to any ambiguity - the parameters should be the same for the same bus on each repeat.  This is inefficient in the dumped state, but not horribly so.

One downside is that once a bus has been mentioned, it gets evaluated from then on, with some CPU cost even if it's not doing anything.  So providing params for all 4 buses in a state dump will (I think) result on "activating" all 4 buses when reloaded.  If a bus isn't being used, I'm not sure we need to include its parameters in a state dump?  

I'm OK with adding the bus value to the first line of the `yield_synth_commands` output so it's set even if `include_fx` is false.  

I'm OK with separate dumps for bus parameters if we limit it to the currently-active buses (0 .. `amy_global.highest_bus`).

Did you encounter a 16kB limit, or is this just an observed risk?  That sounds like an awful lot of synth state.

mateusz28011:
Pushed an update along these lines.

Routing: yes, I actually hit this. The `include_fx=true` line does carry the bus, but
because the event also has FX fields it gets classified as bus-directed on replay — the
bus value selects which bus’s FX to update, and the instrument routing is never touched. Moving
the bus into the first `yield_synth_commands` event fixes it (that line was already there,
commented out). Dropped the separate routing-only line.

Bus dumps are now limited to `0..highest_bus` as you suggested, with a test that inactive
buses stay out.

MicroPython/WASM: WASM is actually my main use case (a browser tool driving AMY for live
editing), so yes — this PR adds public `amy.dump_state()` with the same arbitration as
`get_synth_commands`, and I have the `modtulip.c` wrapper working on the board and in
AMYboard Web. I'll send that as a separate tulipcc PR.

16 KiB: honestly no — my real dumps are ~1 KB. I only reproduced the truncation with a
synthetic 32-instrument state, so I've dropped the dynamic-growth change from this PR to
keep it focused. Can send it separately if you think it's worth having.

(Updated the four `get_synth_commands` golden strings for the bus in the preamble;
`make test` green, 114.)

mateusz28011:
Rebased on latest main and narrowed this to the dump layer only: bus state emitted once per active bus (`0..highest_bus`) plus public `amy.dump_state()`. Updated the PR description to match.

Routing now comes from the synth preamble in #931 (thanks for fixing the root cause there), so I dropped my workaround for it, and I removed the 16 KiB buffer change entirely. No shared files with #931 — the two compose; once it lands, dumps carry routing with nothing extra needed here.

dpwe:
> because the event also has FX fields it gets classified as bus-directed on replay — the bus value selects which bus’s FX to update, and the instrument routing is never touched.

This was a major bug: for `amy.send(synth=1, bus=2, volume=3)` to entirely ignore the `synth` value is syntactically wrong.  I had not properly thought through what happens when events mix osc-, synth-, and bus-directed commands.  I've reworked it and included some tests of possible combinations in #931, I think the commands all now do what I expect them to.

So you should not need an extra `iy1` in a synth command set that includes `y1V1.000` etc., the replay will prefix that line with `i1` (or whatever), and it will set the bus for that synth.  `amy.get_synth_commands(1, include_fx=False)` *will* include `iy` in the prefix if the synth is on a non-default bus (i.e. not bus 0).

mateusz28011:
Confirmed — rebased on main with #931 in and it works exactly as you describe: the dump uses `include_fx=false`, the preamble carries the bus (`i17iv1in1y2Z`), and dump → reset → replay restores routing with nothing extra here. The per-bus `y…` lines only restore FX. Full suite (116) + state-dump tests pass.

dpwe:
I appreciate you finding the problem(s) with the existing `dump_state` and providing working solutions.  I ended up simply copying the bits I wanted to keep and substituting for the bits I wanted to change in #934, but I don't want to discourage you from submitting fixes - it's very helpful!

PR #927, opened 2026-07-18
CPython amy module: bake in the Gamma9001 drum banks

bwhitman:
## Problem

The GM drum kits at patches 384-390 didn't work in the CPython `amy` module (reported by Dan). The kit patch strings ship unconditionally in `patches.h`, but the samples they reference only exist on `-DGAMMA9001` builds (web, Tulip, AMYboard). The CPython build never defined the flag, so pcm.c's preset lookup hit its "out of range → default to preset 0" fallback: the kits loaded but every drum played the wrong sample, and the drums.bin-backed kits (385+) had no blob at all.

## Fix

Mirror the web build:

- **setup.py** generates `build/drums_bin.c` from `sounds/gamma9001/` (via `python -m amy.headers gamma9001`, stdlib-only so it works in pip's isolated build env; skipped gracefully if `sounds/` is absent), links it into the extension, and compiles with `-DGAMMA9001`.
- **pyamy.c** points AMY at the linked-in blob with `amy_set_gamma9001_pcm()` at module init, covering both `amy.live()` and offline rendering.
- Docs updated to list the CPython module among the Gamma9001 platforms.

The wheel grows ~3.7 MB from the baked sample data.

## Test ref changes

`GAMMA9001` also switches the ROM PCM set from `pcm_tiny` to the full 19-sample gamma808 TR-808 bank (by design — the kit note maps index into it), so the python module's PCM presets 0-18 and the channel-10 default drums (patch 384 instead of 258's tiny map) now sound like web/Tulip. The 18 bit-exact refs that render ROM PCM presets or channel-10 drums are regenerated; each was checked to legitimately exercise that content. All 114 tests pass bit-exact against the new refs.

## Verification

All seven kits (384-390) render distinct, non-silent audio from the python repl, including the drums.bin-streamed ones (385+), and `385 != 384` sample data confirmed.

Note: `amy-example` (plain C desktop build) intentionally stays on `pcm_tiny`; happy to switch it in a follow-up if we want desktop C to match.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1181

PR #931, opened 2026-07-18
Set bus and synth

dpwe:
#926 observed that `amy.send(synth=1, bus=1, echo=1)` essentially ignored the `synth` part because the event was classed as "bus-directed".  This is corrected; `amy_event_to_deltas_queue` first handles any bus-directed (global-FX) parts to the event, then goes on to look at osc-directed parts.  This includes passing events with both `bus` and `synth` set to `patches_event_has_voices`, so the synth bus is correctly set.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1186

PR #936, opened 2026-07-19
src/patches.c: Rework how synth properties are pulled from events.

dpwe:
When an event has `synth` set, it is redirected to `patches_event_has_voices` to manage the conversion of synth events into events for individual oscs, including voice management for polyphonic synths.  There used to be a function `patches_voices_for_event` that would wrap either returning all the voices for that synth (e.g. for synth-wide parameter updates), or interfacing with `instruments.c` to figure out voice management for note on/off events.

It also pulled a bunch of synth-directed fields from the event, meaning it had to be called early in the function, before any early exit.  However, this meant it was called before any `NOTES_VIA_MIDI` rewriting of the event, which meant any likely voice management (note on/off tracking, stealing, etc) had already been applied, and would then get applied again when the rewritten event appeared.  

This wasn't noticed because we got lucky - a note-on followed immediately by a repeated note-on had no impact, since repeated notes are routed to the same voice in `instruments.c`.  But note-offs are not like that: the first note-off returned the corresponding voice; the second note-off was ignored since that note was no longer active.  (Also, the note likely changes).

This wasn't noticed because drum synths ignore note-offs anyway.  It was in fact covered in a test, but the ignored note-off was near the end of the sample so I didn't notice the wrong sound. 

The necessary synth-tier extraction has now been split out into a separate function, called early in `patches_event_has_voices`, so that `patches_voices_for_event` is now called only *after* any `NOTES_VIA_MIDI` rewriting, and note-offs are correctly tracked.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1193

PR #943, opened 2026-07-19
Remove 'r'/'voices=', do not expose explicit voice selection to API.

dpwe:
Before we added `synth=`,  we used `voices=` to configure a block of oscs, and to modify several such blocks in parallel.   `synth` still uses the internal `voices` concept, but now we have removed the possibility to use it externally, which was very limited since you never knew which voice was which.  If you want a voice, set up a synth with `num_voices=1`.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1203

PR #949, opened 2026-07-20
Reject `mod_source` cycles so chained modulators can't recurse unbounded [updated]

rt-rtos:
## Update - 25318aa
- **`chained_osc_would_cause_loop` fix**: applied as suggested - the walk now
  calls `ensure_osc_allocd(next_osc)`, matching the osc it's about to
  dereference, instead of re-allocating the head of the chain on every
  iteration.

- **Tests added** to `amy/test.py` with reference wavs:
  - `TestChainedModOsc` uses your tamer version (freq mod depth 0.1,
    `amp={'const': 0.1, 'mod': 0.4}`).
  - `TestModSourceLoopRejected` mirrors the same scene, plus the one send that
    would close the 1<->2 modulator cycle. I updated it to match your version
    of the working test rather than keeping my original deeper-modulation
    settings, to preserve the property that makes it a nice regression test:
    since the rejected send must have no side effects, its output should be
    *identical* to `TestChainedModOsc` - and it is, the two rendered wavs are
    bit-identical. Before the guard, this test recurses until the stack runs
    out.

Verified locally: both new tests pass (err = -100 dB against their refs), the
rejection path logs `osc 1 as mod_source for osc 2 would cause loop.` exactly
once, and the rest of the suite is unchanged.

---

## Summary

`CHAINED_OSC` assignments are already checked for cycles by
`chained_osc_would_cause_loop()`. `MOD_SOURCE` assignments are not, even though
`mod_source` is evaluated recursively at render time. This adds the missing
mirror-image guard, `mod_osc_would_cause_loop()`, so a modulator cycle is
rejected when it is assigned rather than overflowing the stack when it is
rendered.

## Background: modulators can themselves be modulated

An oscillator's `mod_source` is evaluated through
`hold_and_modify()` -> `compute_mod_scale()` -> `hold_and_modify(source)`, which
means a modulator may itself have a `mod_source`. That is a genuinely useful
behaviour.

The same recursion that allows this behaviour can make a cycle fatal.

```mermaid
graph LR
  subgraph valid ["Valid chain (terminates)"]
    direction LR
    A0["osc 0<br/>carrier"] --> A1["osc 1<br/>vibrato LFO"]
    A1 --> A2["osc 2<br/>slow LFO"]
    A2 --> A3(["unset<br/>recursion ends"])
  end

  subgraph cycle ["Cycle (never terminates)"]
    direction LR
    B0["osc 0<br/>carrier"] --> B1["osc 1"]
    B1 --> B2["osc 2"]
    B2 -.->|"closes the loop"| B1
  end
```

## The gap

`compute_mod_scale()` guards only against the one-step self-reference:

```c
uint16_t source = synth[osc]->mod_source;
if(AMY_IS_SET(source)) {
    if(source != osc) {  // that would be weird
        hold_and_modify(source);
        return compute_mod_value(source);
    }
}
```

That catches `a -> a`. It does nothing for `a -> b -> a`, which recurses until
the stack is exhausted.

## The change

`mod_osc_would_cause_loop()` walks the `mod_source` chain from the proposed
modulator and rejects the assignment if it arrives back at the assigning
oscillator. The `MOD_SOURCE` delta handler now only applies its side effects
(allocating the osc, clearing `COEF_VEL`, starting the modulator) when the
assignment is accepted, and unsets `mod_source` otherwise. Structure and
diagnostics deliberately mirror `chained_osc_would_cause_loop()` directly above
it.

The handler also now range-checks `mod_osc` against `AMY_OSCS` before use, which
is why the local changes from `uint16_t` to `int` - an unsigned local would make
the lower bound vacuous.

### One intentional divergence from the chained-osc version

The existing `chained_osc_would_cause_loop()` calls
`ensure_osc_allocd(chained_osc, NULL)` inside its walk loop, but dereferences
`synth[next_osc]`. It therefore only ever allocates the first oscillator in the
chain while reading from every one of them. The new function allocates
`next_osc`, matching what it is about to dereference:

```c
ensure_osc_allocd(next_osc, NULL);
next_osc = synth[next_osc]->mod_source;
```

This is called out because it looks like a transcription slip but is not. Happy
to split the equivalent fix for `chained_osc_would_cause_loop()` into its own
commit, or leave it alone here, whichever is preferred.

## Testing

This PR touches `src/` only. Two test cases are included below rather than in
the diff, since they'd need reference wavs generated before they'd pass and
that's a judgement call about expected output that belongs with the maintainer,
not with this change. Drop them into `amy/test.py` if they look right.

The first exercises the working feature; the second is the regression case, and
before this change it recurses until the stack runs out.

```python
class TestChainedModOsc(AmyTest):
  """A modulator that is itself modulated: slow LFO varies a vibrato LFO's depth."""

  def run(self):
    # osc 2: 0.7 Hz triangle, the "breathing" control signal.
    amy.send(time=0, osc=2, wave=amy.TRIANGLE, freq=0.7, amp=1.0)
    # osc 1: 5 Hz vibrato LFO whose amplitude (i.e. vibrato depth) breathes
    # with osc 2 via its own mod_source.
    amy.send(time=0, osc=1, wave=amy.SINE, freq=5, amp='0.1,0,0,0,0,1', mod_source=2)
    amy.send(time=0, osc=0, wave=amy.SINE, mod_source=1, freq='0,1,0,0,0,1')
    amy.send(time=100, note=70, vel=1)
    amy.send(time=900, vel=0)


class TestModSourceLoopRejected(AmyTest):
  """A mod_source cycle (a->b->a) must be rejected at assignment, not recurse
  unboundedly at render time. The second send below closes a 1<->2 cycle and
  must be dropped; the render then proceeds with osc 2 unmodulated."""

  def run(self):
    amy.send(time=0, osc=2, wave=amy.TRIANGLE, freq=0.7, amp=1.0)
    amy.send(time=0, osc=1, wave=amy.SINE, freq=5, amp='0.1,0,0,0,0,1', mod_source=2)
    amy.send(time=0, osc=2, mod_source=1)   # would close 1<->2: rejected
    amy.send(time=0, osc=0, wave=amy.SINE, mod_source=1, freq='0,1,0,0,0,1')
    amy.send(time=100, note=70, vel=1)
    amy.send(time=900, vel=0)
```


dpwe:
This is great.  Can you make the following changes?

* Fix the bug observed in `chained_osc_would_cause_loop`:
```
@@ -1231,7 +1231,7 @@ int chained_osc_would_cause_loop(uint16_t osc, uint16_t chained_osc) {
     // Check to see if chaining this osc would cause a loop.
     uint16_t next_osc = chained_osc;
     do {
-        ensure_osc_allocd(chained_osc, NULL);
+        ensure_osc_allocd(next_osc, NULL);
         if (next_osc == osc) {
             fprintf(stderr, "chaining osc %d to osc %d would cause loop.\n",
                     chained_osc, osc);
```
* Add the tests as suggested.  But make the "working" test generate a sane output waveform (the suggested one modulates frequency much too deep), e.g.
```
class TestChainedModOsc(AmyTest):
  """A modulator that is itself modulated: slow LFO varies a vibrato LFO's depth."""

  def run(self):
    # osc 2: 1 Hz triangle, the "breathing" control signal.
    amy.send(time=0, osc=2, wave=amy.TRIANGLE, freq=1)
    # osc 1: 5 Hz vibrato LFO whose amplitude (i.e. vibrato depth) breathes
    # with osc 2 via its own mod_source.
    amy.send(time=0, osc=1, wave=amy.SINE, freq=5, amp={'const': 0.1, 'mod': 0.4}, mod_source=2)
    amy.send(time=0, osc=0, wave=amy.SINE, mod_source=1, freq={'mod': 0.1})
    amy.send(time=100, note=70, vel=1)
    amy.send(time=900, vel=0)
```

rt-rtos:
Yes. Will try to get it done tonight. Worst case tomorrow morning 👍

dpwe:
Thanks so much!

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1219

PR #951, opened 2026-07-20
filters.c uses SMUL64R (direct 64x64->64) on desktop and ESP32S3.

dpwe:
In #950, @rt-rtos showed a case where the 32x32 mult wasn't providing enough resolution, and supplied a fix using `int64_t` multiplication.  This turns out to be faster than the `int32_t` multiplication + gymnastics we had been using on ESP32, so I adopted it for all the filtering.

The change leads to numerical differences in all the filter calculations, but they are improvements, particularly the removal of block-floating-point shift-change artifacts.

The block-floating-point path is retained for non-ESP32 MCUs.  I need to test performance on RP2040 and RP2350.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1210

PR #942, opened 2026-07-19
Table-driven cross-platform C API binding codegen

bwhitman:
One function table in `scripts/gen_amy_c_api.py` now generates the AMY C API bindings for every host platform, so `amy.<name>(...)` behaves identically everywhere:

- **CPython**: `src/amy_c_api_py.inc` + `_py_table.inc`, included by a slimmed `pyamy.c` (hand-written wrappers for the 14 table functions deleted; `live`/`start`/`stop`/`config`/`render_to_list`/`inject_midi*` stay hand-written)
- **MicroPython (linked AMY)**: `src/amy_c_api_mp.inc` + `_mp_table.inc`, included by tulipcc's `modtulip.c` (companion PR: shorepine/tulipcc#1201 branch)
- **Web**: `src/amy_c_api.generated.js` (`amy_c_api_bind(module)` + the `AMY_C_API_PY_INSTALL` MicroPython install snippet) and `src/amy_c_api_exports.mk` feeding `EXPORTED_FUNCTIONS` so a table entry can never miss the WASM export list. `amy_connector.js` drops its hand cwrap copy and exposes `amy_c_api` on globalThis.
- **Godot**: generated decl/impl/bind `.inc`s in `godot/src` (native GDExtension) plus a generated GDScript block in `godot/amy.gd` (native `_synth.call` / web `JavaScriptBridge.eval` dispatch)
- **Python dispatch**: a generated `BEGIN/END GENERATED` block in `amy/__init__.py` resolves each `amy.<name>` to `c_amy` → `tulip.amy_*` → JS-installed backend

New API surface gained by generation: `bleep`, `reset_sysclock`, `sequencer_ticks`, `process_single_midi_byte`, `get_output_buffer`, `get_input_buffer` on CPython; `amy.dump_state()` on MicroPython (previously a latent NameError); `render_load`/`set_render_load_threshold`/`dump_state` etc. on web; `render_load`/`dump_state`/`bleep`/`sequencer_ticks`/`reset_sysclock` on Godot.

Workflow: edit the table → `make c-api` → commit. `make check-c-api` runs in CI (new `c-api` job) so drift fails the build. Generated files are committed (the pip build and tulipcc builds consume them directly).

Verified: `make test` (117 pass), CPython smoke of the full surface, Tulip Desktop macOS build + stdin-harness test of `amy.*`, Tulip Web + AMYboard Web in-browser python tests, Godot GDExtension macos/arm64 scons build.

[...]

PR #953, opened 2026-07-20
Commit web JS codegen outputs in src/, verified by check-c-api

bwhitman:
## Why

`amy_api.generated.js` and `patches.generated.js` were generated into the gitignored `build/` by `make web`, so consumers that commit copies (tulipcc's amyboardweb `static/`) could silently drift from their pinned amy — that's exactly what happened in shorepine/tulipcc#1208 after #943 removed `voices`.

## What

- Move `gen_amy_js_api.py` and `gen_patches_js.py` outputs to **committed** `src/amy_api.generated.js` / `src/patches.generated.js` (alongside `amy_c_api.generated.js` from #942), and give both scripts a `--check` mode.
- New `scripts/gen_pcm_presets_js.py` → committed `src/pcm_presets.generated.js`: the web editor's wave-preset table (`window.AMY_WAVE_PRESETS`) from `pcm_tiny.h`. This file previously existed only as a one-off committed in tulipcc with **no generator anywhere**; output is byte-identical to it apart from the header comment.
- `make c-api` runs all the generators; `make check-c-api` (already run by CI on every PR) now fails if any of the three JS outputs is stale.
- `web`/`deploy-web` use the committed files; the `build/` rules are gone.

## Testing

- `make check-c-api` passes; `make test` passes (117 tests).
- tulipcc side (PR to follow): amyboardweb stops committing `static/*.generated.js` copies and copies these from `src/` at build time, so a bare pin bump can never drift them again.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1212

PR #961, opened 2026-07-23
survive out-of-memory on the voice/event allocation paths

rt-rtos:
# survive out-of-memory on the voice/event allocation paths

Don't crash when runtime allocations fail: guard alloc_osc, instrument_init, the delta pool, and midi yield state

## Summary

Several allocations that run during normal playing - not just at startup -
write through their `malloc_caps()` result without checking for NULL. On
embedded targets the heap genuinely runs out (on ESP32-S3, AMY's per-osc
structs live in the scarce ~300 KB internal SRAM region alongside WiFi/BT and
app data), and each of these turns heap exhaustion into a hard fault. Found
while auditing an ESP32-S3 project where dynamically growing a synth's
`num_voices` makes per-osc allocation user-drivable at runtime.

Four families, each now degrading gracefully instead of crashing:

| allocation | on failure now |
|---|---|
| `alloc_osc` (per-osc synthinfo + msynth + breakpoints) | osc left NULL, voice silent, stderr log. Guards at every deref funnel: `play_delta` entry drops the delta; the breakpoint realloc allocates the replacement block before releasing the old one, so a failed grow leaves the osc's existing state intact and only the delta that wanted more breakpoints is dropped; `chained_osc_would_cause_loop` / `mod_osc_would_cause_loop` refuse the link; the `ALGO_SOURCE` branch unsets a source that failed to allocate; `config_chorus` turns the chorus off (mirroring `config_reverb`'s failure path) and the render loop skips a bus whose chorus LFO osc is NULL; `set_cv_from_osc` unsets the CV mapping; PARTIALS note-on skips a NULL partial (note-off already tolerated the gap), INTERP_PARTIALS note-on drops the note |
| `alloc_chorus_delay_lines` (`delay_mod` block) | counted as an allocation failure like the delay lines themselves, so a partial chorus alloc is torn down instead of leaving a NULL `delay_mod` for the render loop's `bzero` |
| `instrument_init` / `voice_fifo_init` (per-synth) | returns NULL, which every caller already treats as "synth not defined" (`instrument_number_exists`) |
| `deltas_pool_alloc` / `deltas_add_pool_block` / `delta_get` (event pool growth) | pool keeps its existing chain, event dropped, stderr log; `add_delta_to_queue` undoes its qsize bump |
| `yield_midi_message_handler_events` yield state | midi message dropped |

The existing NULL tolerance in `reset_osc` and the render loop
(`synth[osc] != NULL && ...`) is what makes the NULL-osc representation safe;
this change extends the same convention to the paths that create oscs.
Running out of osc *slots* (`max_oscs`) was already handled gracefully
("cannot find N oscs ... not setting this voice"); running out of *heap* for
an osc was not.

Boot-time allocations in `amy_start()` (osc pointer tables, output blocks,
buses) are deliberately untouched - failing those means the process can't run
at all, and they don't recur.

Reworking the breakpoint realloc surfaced a **pre-existing out-of-bounds
write on that path, present on main today and unrelated to OOM** - no
allocation has to fail to hit it. The sizing loop seeded *both* breakpoint
sets at `DEFAULT_NUM_BREAKPOINTS` and only overwrote the one being grown, so
growing one set silently resized the other back to 8 while the copy that
follows still wrote its old number of entries into it. Any osc with two
envelopes where one has been grown past 8 breakpoints and the other is grown
later hits it. Reproducer, ASan trace and the one-line fix are at the bottom.

Every failure path reports through one `amy_oom()` helper: release builds log
a line and degrade at the call site, and under `AMY_DEBUG` it aborts there
instead, so a developer who wants the crash still gets a backtrace at the
cause. It also counts the failures, exposed as `amy_get_oom_count()` - a
degraded synth is otherwise indistinguishable from one playing exactly what
it was asked to, and the host needs some signal to decide to re-create it.
Same shape as the existing overload counter and `amy_get_render_load()`.

### Verification

- `make test`: the pass/fail set is byte-identical to pristine main in the
  same environment (the guards are added branches only; no behavior change
  when allocation succeeds).
- OOM harness (standalone, not part of the PR): starts AMY headless with
  `max_oscs = 4000`, proves a small synth plays, then caps `RLIMIT_AS`
  32-512 KB above current usage and keeps creating 24-voice synths and
  playing notes on them. It finishes by growing an existing osc's envelope
  from 8 to 12 breakpoints while the heap is still exhausted, which exercises
  the realloc path (the osc has to keep its old, smaller vectors), and it
  asserts on `amy_get_oom_count()` at both ends: zero before the rlimit, and
  non-zero at the end, so the run can't pass by never failing an allocation.
  - pristine main: SIGSEGV at every headroom tried (first in `alloc_osc`,
    then `instrument_init`, then the delta pool / midi yield state as each
    earlier site is guarded)
  - with this PR: stderr reports the failures
    (`alloc_osc: out of memory allocating osc N`, `synth N not defined`),
    audio keeps rendering, clean exit - at every headroom tried. The final
    `oom_count` matches the number of reported failures exactly (1058 in the
    run below).

Harness source (drop in repo root, build after `make amy-example` per the
comment at the top of the file): `oom-repro.c` below along with truncated log output; to avoid polluting git history.

### Notes for reviewers

- `play_delta`'s entry guard exempts `RESET_OSC` (matching the existing
  `ensure_osc_allocd` exemption); `reset_osc()` already tolerates NULL.
- The loop-helper guards return "would cause loop" for an unallocatable
  target; both callers' failure branches (refuse the chain, unset
  mod_source) are exactly what should happen for a link to an osc that
  doesn't exist.
- `voice_fifo_free(NULL)` is `free(NULL)`, so the partial-failure cleanup in
  `instrument_init` is safe.
- The sweep covers every `ensure_osc_allocd` caller: the sites that
  dereference `synth[osc]` immediately afterwards (chorus setup and render,
  `cv_trigger.c`, `interp_partials.c`) now check for NULL; the two callers in
  the event-to-delta path deliberately don't, because they don't dereference
  locally - the deref happens later in the guarded `play_delta`.

  ---

# oom-repro.c and output 

Test harness:

```c
// oom-repro.c - demonstrate AMY behavior when the heap runs out while
// allocating oscs (see the alloc_osc guards in src/amy.c).
//
// Build after `make amy-example`:
//   gcc -O2 -Isrc oom-repro.c src/algorithms.o src/amy.o src/envelope.o \
//     src/examples.o src/parse.o src/filters.o src/oscillators.o src/pcm.o \
//     src/interp_partials.o src/custom.o src/delay.o src/log2_exp2.o \
//     src/patches.o src/transfer.o src/sequencer.o src/libminiaudio-audio.o \
//     src/instrument.o src/amy_midi.o src/api.o src/midi_mappings.o \
//     src/cv_trigger.o -lm -pthread -o oom-repro
//
// It starts AMY headless, proves a small synth works, then caps the address
// space (RLIMIT_AS) just above current usage and keeps allocating 24-voice
// synths until malloc fails, and finally grows an existing osc's envelope
// while the heap is still exhausted.
//   - without the guards: segfault inside alloc_osc / play_delta
//   - with the guards:   "alloc_osc: out of memory..." on stderr, then
//                        "SURVIVED heap exhaustion" and exit 0
//
// amy_get_oom_count() only exists with the guards applied; drop the two
// assertions on it to build this against pristine main for comparison.
#include "amy.h"
#include <stdio.h>
#include <sys/resource.h>
#include <unistd.h>
#include <signal.h>
#include <execinfo.h>
#include <stdlib.h>

// examples.o expects the embedding app to provide this.
void delay_ms(uint32_t ms) { (void)ms; }

static void segv_handler(int sig) {
    void *frames[32];
    int n = backtrace(frames, 32);
    backtrace_symbols_fd(frames, n, 2);
    _exit(139);
    (void)sig;
}

static long vsz_bytes(void) {
    long pages = 0;
    FILE *f = fopen("/proc/self/statm", "r");
    if (f == NULL) return -1;
    if (fscanf(f, "%ld", &pages) != 1) pages = 0;
    fclose(f);
    return pages * sysconf(_SC_PAGESIZE);
}

int main(void) {
    signal(SIGSEGV, segv_handler);
    // Prime backtrace()'s lazy libgcc load while allocation still works, so
    // the handler stays malloc-free under the rlimit.
    { void *f[4]; backtrace(f, 4); }
    amy_config_t cfg = amy_default_config();
    cfg.features.default_synths = 0;
    // Plenty of osc slots so the heap - not max_oscs - is what runs out.
    cfg.max_oscs = 4000;
    amy_start(cfg);

    // Baseline: a small synth allocates and plays fine.
    amy_event e = amy_default_event();
    e.synth = 1;
    e.patch_number = 130;
    e.num_voices = 2;
    amy_add_event(&e);
    for (int i = 0; i < 10; i++) amy_simple_fill_buffer();
    e = amy_default_event();
    e.synth = 1;
    e.midi_note = 60;
    e.velocity = 1.0f;
    amy_add_event(&e);
    for (int i = 0; i < 10; i++) amy_simple_fill_buffer();

    // Baseline raw osc for the breakpoint-grow test below: allocated (with
    // default 8-slot envelope vectors) while memory is still available.
    e = amy_default_event();
    e.osc = 400;
    e.wave = SINE;
    e.freq_coefs[COEF_CONST] = 220.0f;
    amy_add_event(&e);
    for (int i = 0; i < 4; i++) amy_simple_fill_buffer();
    fprintf(stderr, "baseline synth ok (oom_count %u)\n", amy_get_oom_count());
    if (amy_get_oom_count() != 0) return 3;

    // Cap the address space a little above current usage so heap growth
    // fails. OOM_HEADROOM_KB (default 256) varies where the failures land.
    long headroom_kb = 256;
    if (getenv("OOM_HEADROOM_KB")) headroom_kb = atol(getenv("OOM_HEADROOM_KB"));
    long now = vsz_bytes();
    if (now <= 0) { fprintf(stderr, "statm read failed\n"); return 2; }
    struct rlimit rl;
    rl.rlim_cur = rl.rlim_max = (rlim_t)now + ((rlim_t)headroom_kb << 10);
    if (setrlimit(RLIMIT_AS, &rl) != 0) { perror("setrlimit"); return 2; }

    // Keep allocating big synths and playing notes on them until mallocs
    // start failing; render between events so deltas get applied.
    for (int s = 2; s < 16; s++) {
        amy_event big = amy_default_event();
        big.synth = s;
        big.patch_number = 130;
        big.num_voices = 24;
        amy_add_event(&big);
        for (int i = 0; i < 4; i++) amy_simple_fill_buffer();

        amy_event n = amy_default_event();
        n.synth = s;
        n.midi_note = 48 + s;
        n.velocity = 1.0f;
        amy_add_event(&n);
        for (int i = 0; i < 4; i++) amy_simple_fill_buffer();
    }

    // Breakpoint grow under OOM: osc 400 exists with 8-slot envelopes, and a
    // 12-breakpoint envelope makes ensure_osc_allocd grow them. The osc must
    // keep its old vectors when that allocation fails; the guarded build
    // drops the out-of-range breakpoints instead of writing past the arrays
    // (or, previously, crashing on the freed-then-unallocatable osc).
    e = amy_default_event();
    e.osc = 400;
    for (int i = 0; i < 12; i++) {
        e.eg0_times[i] = 10 * (i + 1);
        e.eg0_values[i] = 1.0f / (i + 1);
    }
    amy_add_event(&e);
    for (int i = 0; i < 4; i++) amy_simple_fill_buffer();
    fprintf(stderr, "breakpoint grow under OOM ok\n");

    // The host's signal that any of the above degraded rather than played.
    if (amy_get_oom_count() == 0) {
        fprintf(stderr, "oom_count never moved - no allocation actually failed\n");
        return 3;
    }
    fprintf(stderr, "SURVIVED heap exhaustion (oom_count %u)\n", amy_get_oom_count());
    return 0;
}
```

 ## oom-repro-pristine output `SIGSEGV | _exit(139)`

The pristine comparison build is the same file with the two
`amy_get_oom_count()` assertions dropped - that symbol only exists with this
PR applied. It crashes in phase 2 either way.

```
baseline synth ok
./oom-repro-pristine(+0x171df)[0x5e07bc3691df]
/lib/x86_64-linux-gnu/libc.so.6(+0x45330)[0x78cd8c445330]
./oom-repro-pristine(alloc_osc+0x72)[0x5e07bc36ac72]
./oom-repro-pristine(amy_event_to_deltas_queue+0x35a)[0x5e07bc36e05a]
./oom-repro-pristine(patches_load_patch+0x8a1)[0x5e07bc385611]
./oom-repro-pristine(amy_event_to_deltas_queue+0x427)[0x5e07bc36e127]
./oom-repro-pristine(main+0x33b)[0x5e07bc368fcb]
/lib/x86_64-linux-gnu/libc.so.6(+0x2a1ca)[0x78cd8c42a1ca]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x8b)[0x78cd8c42a28b]
./oom-repro-pristine(_start+0x25)[0x5e07bc3690e5]
```

## oom-repro output - guarded build (fix/alloc-osc-oom-guards)

Same scenario as the pristine run (which segfaults in `alloc_osc`): baseline
synth, then `RLIMIT_AS` capped ~256 KB above current usage, then a loop of
24-voice patch-130 synths, then a breakpoint grow on an osc that already
exists. Log condensed; verbatim lines in code blocks, removed/repeated output
described in prose.

### Phase 1 - baseline (before the rlimit)

A 2-voice synth plays, and a raw osc is allocated with the default 8-slot
envelope vectors for the grow test in phase 4. Nothing has failed yet, and
the counter says so:

```
baseline synth ok (oom_count 0)
```

### Phase 2 - heap exhaustion: the guard fires

Synths 2 and 3 get their voice bookkeeping (cheap table writes succeed), but
the per-osc `malloc` in `alloc_osc` starts failing. Each failure now prints
one line and leaves the osc silent instead of writing through a NULL
`synth[osc]` - this is exactly where the pristine build crashes:

```
alloc_osc: out of memory allocating osc 175
alloc_osc: out of memory allocating osc 168
alloc_osc: out of memory allocating osc 2178
alloc_osc: out of memory allocating osc 2177
```

*[Thousands of further `alloc_osc: out of memory` lines removed. The volume
is expected: every delta that touches a still-unallocated osc goes through
`ensure_osc_allocd`, re-attempts the malloc, and prints again - a patch load
sends many parameters per osc, so each silent osc is reported many times.
Each touch is a fresh retry, so an osc recovers automatically if memory
frees up later.]*

### Phase 3 - synths 4..15: upstream voice cap, not the guard

From synth 4 onward the failure mode changes to a pre-existing, already-safe
upstream path: default `max_voices` is 64 (`src/api.c`), synths 1–3 hold
2+24+24 = 50 voices, so a further 24-voice request is rejected in
`patches_voices_for_load` (`src/patches.c`). One iteration, verbatim:

```
ran out of voices allocating 24 voices to synth 4, ignoring.
synth 4 not defined (get_flags)
synth 4 not defined (get_flags)
synth 4 not defined (get_flags)
synth 4 not defined (get_flags)
synth 4 not defined (voice_for_event)
synth 4 not defined (noteon_delay)
```

*[Removed here, and from every iteration: the `patches_debug()` dump that
upstream emits on this rejection path (it also runs into the "ignoring."
message, which lacks a trailing newline). Each dump listed all 50 voices
with their base oscs - even voices based at 2000+, odd at 0+, because AMY
splits the osc pool in half across its two render cores and round-robins
voices between them; bases step by 8 (patch 130 uses 8 oscs per voice) -
followed by the full 4000-entry `osc_to_voice` table, mostly 255
(= unassigned), plus the synth table showing synths 1–3 intact. The dump was
byte-identical in all 12 iterations: the voice/osc tables stayed uncorrupted
through thousands of failed allocations.]*

*[Synths 5 through 15 repeat the identical pattern - `ran out of voices ...
ignoring.`, the unchanged `patches_debug()` dump, and six `synth N not
defined` lines for the follow-up note-on.]*

### Phase 4 - growing an existing osc's breakpoints while out of memory

Osc 400, allocated in phase 1 with 8-slot envelope vectors, is sent a
12-breakpoint envelope. `ensure_osc_allocd` has to move it to a bigger block,
which fails: the replacement is allocated before the old block is released,
so the osc keeps its existing 8-slot vectors and its parameters, and only the
breakpoints past slot 8 are dropped. Verbatim (one line per delta in the
event that touched the osc):

```
alloc_osc: out of memory allocating osc 400
alloc_osc: out of memory allocating osc 400
alloc_osc: out of memory allocating osc 400
alloc_osc: out of memory allocating osc 400
alloc_osc: out of memory allocating osc 400
alloc_osc: out of memory allocating osc 400
alloc_osc: out of memory allocating osc 400
alloc_osc: out of memory allocating osc 400
breakpoint grow under OOM ok
```

## Result

```
SURVIVED heap exhaustion (oom_count 1058)
```

Exit 0, and 1058 is also the exact number of `out of memory` lines in the
full log - every guarded failure is counted once, and a host polling
`amy_get_oom_count()` sees the same events the log shows. The pristine binary
segfaults in phase 2; the guarded build degrades to silent oscs and
completes. Note the genuine heap-OOM stress is concentrated in synths 2–3 -
later iterations exercise the voice cap instead. To make every iteration
hammer `alloc_osc`, set `cfg.max_voices = 1000;` in `oom-repro.c`.

---

# bp-grow.c - the pre-existing breakpoint realloc bug

Independent of OOM: nothing here runs out of memory. Two envelopes on one
osc, one of them grown past `DEFAULT_NUM_BREAKPOINTS` before the other is
grown, is enough.

```c
// bp-grow.c - the breakpoint-realloc sizing bug in ensure_osc_allocd, which is
// pre-existing on main and independent of the OOM guards (no allocation fails
// here). Growing one breakpoint set resized the *other* one back to
// DEFAULT_NUM_BREAKPOINTS, while the copy that follows still wrote the old,
// larger number of entries into it.
//
// Build after `make amy-example`, ideally with ASan:
//   gcc -g -O1 -fsanitize=address -Isrc bp-grow.c src/algorithms.o src/amy.o \
//     src/envelope.o src/examples.o src/parse.o src/filters.o \
//     src/oscillators.o src/pcm.o src/interp_partials.o src/custom.o \
//     src/delay.o src/log2_exp2.o src/patches.o src/transfer.o \
//     src/sequencer.o src/libminiaudio-audio.o src/instrument.o \
//     src/amy_midi.o src/api.o src/midi_mappings.o src/cv_trigger.o \
//     -lm -pthread -o bp-grow
//   ASAN_OPTIONS=detect_leaks=0 ./bp-grow
//
// Case A (grow set 1, set 0 already at 16) corrupts set 1's vectors, but the
// overrun stays inside the osc's own malloc block, so it is silent.
// Case B (grow set 0, set 1 already at 16) overruns the last vectors in the
// block and runs off the end of the allocation - ASan reports it.
#include "amy.h"
#include <stdio.h>

void delay_ms(uint32_t ms) { (void)ms; }

// Give the osc `n` breakpoints in set `set`, then let the deltas apply.
static void set_envelope(int osc, int set, int n) {
    amy_event e = amy_default_event();
    e.osc = osc;
    for (int i = 0; i < n; i++) {
        if (set == 0) {
            e.eg0_times[i] = 10 * (i + 1);
            e.eg0_values[i] = 1.0f / (i + 1);
        } else {
            e.eg1_times[i] = 10 * (i + 1);
            e.eg1_values[i] = 1.0f / (i + 1);
        }
    }
    amy_add_event(&e);
    for (int i = 0; i < 4; i++) amy_simple_fill_buffer();
}

static void sizes(const char *when, int osc) {
    printf("%-24s osc %d max_num_breakpoints = %d, %d\n", when, osc,
           synth[osc]->max_num_breakpoints[0], synth[osc]->max_num_breakpoints[1]);
}

int main(void) {
    setvbuf(stdout, NULL, _IONBF, 0);  // keep the log if case B aborts under ASan
    amy_config_t cfg = amy_default_config();
    cfg.features.default_synths = 0;
    amy_start(cfg);

    // Case A: set 0 at 16 slots, then grow set 1. Set 0 must keep its 16.
    set_envelope(100, 0, 16);
    sizes("A after eg0 x16:", 100);
    set_envelope(100, 1, 12);
    sizes("A after eg1 x12:", 100);
    int bad = synth[100]->max_num_breakpoints[0] < 16;
    if (bad) printf("  FAIL: set 0 shrank to %d with 16 entries copied into it\n",
                    synth[100]->max_num_breakpoints[0]);
    for (int i = 0; i < 16 && !bad; i++) {
        uint32_t want = ms_to_samples(10 * (i + 1));  // times are stored in samples
        if (synth[100]->breakpoint_times[0][i] != want) {
            printf("  FAIL: set 0 entry %d is %u, expected %u\n",
                   i, synth[100]->breakpoint_times[0][i], want);
            bad = 1;
        }
    }
    if (!bad) printf("  OK: set 0 kept its size and contents\n");

    // Case B: the same thing with the sets swapped. Set 1's vectors are last
    // in the osc's block, so the overrun leaves the allocation entirely.
    set_envelope(200, 1, 16);
    sizes("B after eg1 x16:", 200);
    set_envelope(200, 0, 12);
    sizes("B after eg0 x12:", 200);
    if (synth[200]->max_num_breakpoints[1] < 16) {
        printf("  FAIL: set 1 shrank to %d with 16 entries copied into it\n",
               synth[200]->max_num_breakpoints[1]);
        bad = 1;
    } else {
        printf("  OK: set 1 kept its size\n");
    }
    return bad;
}
```

## Output on main (upstream @ 3555365, built with `-fsanitize=address`)

Case A shows the silent form - set 0 is resized 16 -> 8 and then written with
16 entries, which stays inside the osc's own block and just corrupts the
neighbouring vectors. Case B is the same bug on the last vectors in the
block, where it leaves the allocation:

```
A after eg0 x16:         osc 100 max_num_breakpoints = 16, 8
A after eg1 x12:         osc 100 max_num_breakpoints = 8, 16
  FAIL: set 0 shrank to 8 with 16 entries copied into it
B after eg1 x16:         osc 200 max_num_breakpoints = 8, 16
=================================================================
==93109==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x5160000017f0 at pc 0x59597822e057 bp 0x7fff5f57bca0 sp 0x7fff5f57bc90
WRITE of size 4 at 0x5160000017f0 thread T0
    #0 0x59597822e056 in ensure_osc_allocd src/amy.c:1004
    #1 0x5959782309b9 in play_delta src/amy.c:1349
    #2 0x5959782375c8 in amy_execute_deltas src/amy.c:1983
    #3 0x595978306134 in amy_simple_fill_buffer src/api.c:229
    #4 0x59597822932d in set_envelope bp-grow.c:40
    #5 0x5959782298a2 in main bp-grow.c:76
    #6 0x70f3a7e2a1c9  (/lib/x86_64-linux-gnu/libc.so.6+0x2a1c9) (BuildId: 8e9fd827446c24067541ac5390e6f527fb5947bb)
    #7 0x70f3a7e2a28a in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2a28a) (BuildId: 8e9fd827446c24067541ac5390e6f527fb5947bb)
    #8 0x595978228fc4 in _start (./bp-grow-base+0x27fc4) (BuildId: 3ed93c42d6585383cb9e8e8b74f3785d42efda64)

0x5160000017f0 is located 0 bytes after 624-byte region [0x516000001580,0x5160000017f0)
allocated by thread T0 here:
    #0 0x70f3a82fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x59597822d152 in malloc_caps src/amy.c:196
    #2 0x59597822d152 in alloc_osc src/amy.c:925

SUMMARY: AddressSanitizer: heap-buffer-overflow src/amy.c:1004 in ensure_osc_allocd
```

## Output with this PR

```
A after eg0 x16:         osc 100 max_num_breakpoints = 16, 8
A after eg1 x12:         osc 100 max_num_breakpoints = 16, 16
  OK: set 0 kept its size and contents
B after eg1 x16:         osc 200 max_num_breakpoints = 8, 16
B after eg0 x12:         osc 200 max_num_breakpoints = 16, 16
  OK: set 1 kept its size
```

Exit 0, ASan clean. The fix is the seed in that sizing loop - start from the
set's current size instead of `DEFAULT_NUM_BREAKPOINTS`, so a set is never
shrunk out from under the copy that follows:

```c
// Never shrink a set; the copy below has to fit its old contents.
new_max_num_breakpoints[i] = synth[osc]->max_num_breakpoints[i];
```

It rides along in the realloc rework because that rework rewrote the same
loop, but it is a separate defect and can be split into its own commit if
you'd rather review it that way.


rt-rtos:
We may want a macro toggle? This seems like a typical prod/debug split. Crash as a diagnostic is useful if you are capable of acting on it, but the signal loses its value in "prod" where it would be better handled gracefully?

A new flag or under one of the existing diagnostics flags if so?

dpwe:
Thanks again for this clean PR.

I didn't bother with checking for OOM on the reasoning that once you run out of RAM, everything's going to be messed up anyway.  What's the scenario when you end up with some non-allocated oscs, but you keep on doing useful work?  I kind of assumed the engine would be in a difficult-to-know state, so it's game over anyway.

Clean exception handling is always nice, but putting a lot of guards in the code make it harder to read.  That's my only reservation.

rt-rtos:
Edit: I missed the part of your reply expressing concern over the guards. So I may have ran away with tunnelvision on the curiosities you expressed rather than you actual concern. 
Feedback welcome on what form would be easiest to follow (if this is worth pursuing). I wouldn´t mind doing another rework.

---

**Updated - Original PR**
- Updates to original PR body
- Rebased onto upstream main (1.2.89)
- New non-OOM breakpoint reproducer and ASan trace

---

**Regarding feedback and post-feedback PR changes:**

Thanks - fair question, and it sent me back into the code. Short version: the
crash-vs-degrade difference only matters when AMY is a component rather than
the whole device, the post-OOM state is better-defined than I had assumed
(with two exceptions I found and fixed below), and I made an effort to
improve readability throughout. I believe this ultimately comes down to a
judgement call on what is / isn't the job of the AMY engine, and that call is
entirely in your court.

**The scenario.** When AMY basically is the application, I agree with you - if
allocation fails the device is probably in trouble and a crash isn't much
worse. In cases like mine, where AMY is the core component in a larger
multi-component firmware, heap pressure usually isn't AMY's doing and is
often temporary: WiFi buffers, a display stack, the host app briefly holding
internal SRAM, or fragmentation blocking one contiguous alloc while total
free heap looks fine.

Growing `num_voices` at runtime turns that into a hard fault and a reboot
mid-performance; with the guards it's a silent voice and a stderr line. Same
user-visible event as running out of osc *slots*, which is already handled
gracefully ("cannot find N oscs ... not setting this voice") - it just
arrives via the heap instead of `max_oscs`.

Could the app handle it outside the engine? I checked - `malloc_caps` is
overridable via `MALLOC_CAPS_DEFINED`, so the app controls where to allocate,
but it can't control what AMY does when the answer is NULL. The recovery
choice (leave the osc NULL, drop the delta, tear down the half-built chorus)
depends on which invariants are half-established at that moment, and only the
engine knows that. Reserve pools and free-heap checks can make the failure
rarer or later, but the allocs are lazy and fire mid-render, so a pre-flight
check in the app can't cover them.

#### How knowable is the post-failure state? Two layers:

- *Memory safety*: knowable by construction. An osc is one malloc (synthinfo
  + msynth + breakpoint vectors), so `synth[osc]` is either NULL or fully
  initialized - no partial osc exists, and NULL-and-skipped is already the
  resting state of every untouched osc. The delta pool and `instrument_init`
  are all-or-nothing too (event dropped whole, no half-registered synth). The
  harness agrees: `patches_debug()` dumps stayed byte-identical across
  iterations after thousands of failed allocs.
- *Musical state*: you have a point here. The failure states are enumerable
  (silent voice, refused chain/mod link, chorus off) but not queryable, and
  retry-on-touch has one awkward case - memory freeing up mid-patch-load
  gives an osc reset defaults plus only the tail of the patch's params.
  Bounded, but audibly wrong rather than silent.

A crash is a useful diagnostic signal, and where AMY is the whole device it
may well be the one you want. On the knowability question specifically I'd
call it a wash: crashing doesn't tell the app more about the state AMY was
in, it just ends that state. And as far as I can tell each guarded state is
recoverable through the ordinary paths - `reset_osc`, a patch reload, or
re-creating the synth all work normally afterward.

What was missing was the host's ability to *notice*, so I added it: a new
commit exposes `amy_get_oom_count()`, a count of runtime allocation failures
since `amy_start`. The app doesn't need to know which osc failed, just that
one did, and re-creating the synth is the known-good reset. Every guard
already reports through a single `amy_oom()` helper, so it's one increment
and a getter (+17/-3), the same shape as the existing overload counter and
`amy_get_render_load()`. It's the last commit on the branch and nothing
before it depends on it, so it drops cleanly if you'd rather not have the
API surface - say the word and I'll force-push without it.

**A correction, found while writing the above.** The breakpoint-realloc path
had a window my harness never exercised: it freed the old osc before
allocating the bigger one, and my re-check sat in the caller - after a deref
inside `ensure_osc_allocd` would already have happened. Confirmed against the
previous head of this branch (harness now grows a 12-breakpoint envelope
under OOM; patch 130 never needs more than the default 8, which is why it was
missed):

```
$ addr2line -f -e oom-repro-prev 0x8045 0xa89c
ensure_osc_allocd
play_delta
```

Fixed by allocating the new block before freeing the old, so on OOM the osc
keeps its previous state and only the delta that wanted more breakpoints is
dropped.

**A pre-existing out-of-bounds write on the same path, unrelated to OOM.**
The realloc sizing loop seeded *every* breakpoint set at
`DEFAULT_NUM_BREAKPOINTS` and only overwrote the one being grown, so growing
one set silently resized the other back to 8 while the copy that follows
still wrote its old number of entries into it. No allocation has to fail to
hit this - it reproduces on main (`3555365`) with a plain two-envelope osc.
Growing set 0 while set 1 already sits at 16 is the bad direction: set 1's
vectors are last in the osc's block, so the overrun leaves the allocation
entirely.

```
ERROR: AddressSanitizer: heap-buffer-overflow ... WRITE of size 4
    #0 ensure_osc_allocd src/amy.c:1004
    #1 play_delta src/amy.c:1349
0x5160000017f0 is located 0 bytes after 624-byte region allocated by alloc_osc
```

(The other direction - growing set 1 while set 0 sits at 16 - corrupts the
neighbouring vectors inside the osc's own block, so it's silent.) The rework
never shrinks a set, which closes both. Reproducer and full trace are in the
PR body under `bp-grow.c`.

This one is independent of the OOM argument, so I'll open it as its own PR
regardless of how this one lands.

**Readability.** Took the point and made improvements:

- `ensure_osc_allocd` now returns bool ("allocated, with the capacity you
  asked for"), so every ensure-then-peek pair collapses to
  `if (!ensure_osc_allocd(...))` - one line per guard, and the same check
  fixes a matching capacity hazard in `interp_partials.c` for free.
- Every failure path reports through one `amy_oom()` helper: release builds
  log one line and degrade, under `AMY_DEBUG` it aborts at the failure site -
  crash-as-diagnostic where you can act on it, graceful where you can't (the
  toggle from my earlier comment, verified both ways).
- The rework commit is +86/-78: the guarded version ends up slightly smaller
  than the save/restore dance it replaced.

`make test` pass/fail set unchanged throughout, compared against the same
upstream base.

P.S. - thanks for taking the time on this one, and for the feedback; it made the patch better. Equally happy to close it if you decide it's outside the intended scope - I honestly don't know how much an embedder is expected to rearrange AMY's internals; That´s a question of intent that only you/Brian can answer.

dpwe:
Thanks, I take your point.  Blindly segfaulting is really not polite behavior.

> Fixed by allocating the new block before freeing the old, so on OOM the osc keeps its previous state and only the delta that wanted more breakpoints is dropped.

This makes me nervous, since it guarantees heap fragmentation - those subsequently-freed smaller blocks are never going to get reused.  I think I'd rather risk the catastrophic mid-flight malloc fail than effectively plant a memory leak.  

But good catch about the inadvertent resize-down of the other BP set, if I'm understanding right.

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1232

rt-rtos:
Thanks.

On fragmentation - right to be vigilant, and the peak cost is real: the grow briefly holds both blocks, so alloc-first can fail where free-first would have squeaked through. I tried to construct the leak, though, and to the best of my understanding it doesn't materialize: the freed hole is an exact fit for the next default-size `alloc_osc`, and with live oscs on both sides free-first ends in the same place - nothing to coalesce with, the bigger block lands elsewhere, same hole. If you have a scenario in mind I'm happy to run it through the harness.

The transient peak could avoided: keep free-before-alloc, and on failure NULL the osc into the already-handled silent-voice case. No peak increase, still no crash; the cost is losing that osc's state on a failed grow. Happy to send that as a small follow-up if you prefer the trade.

And yes, exactly right on the BP-set bug. The never-shrink logic in this PR closes both directions. `bp-grow.c` reproducer under ASan comes back clean on merged main (1.2.92).

Issue #966, opened 2026-07-24
Out-of-bounds write in patches_load_patch()'s oscillator allocator corrupts unrelated voices


substatica:
# Out-of-bounds write in `patches_load_patch()`'s oscillator allocator corrupts unrelated voices

## Summary

`patches_load_patch()` in `src/patches.c` searches for a free block of `oscs_per_voice` consecutive oscillators when loading a patch onto a synth. The search wraps the **candidate start index** around the end of the oscillator pool, but never wraps the **offsets within the candidate block**. When a candidate block straddles the end of the pool, the code reads and writes past the end of `osc_to_voice[]` — directly into `voice_to_base_osc[]`, which is allocated immediately afterward — corrupting the voice-ownership bookkeeping for a completely unrelated synth/voice.

**Symptom:** loading or changing a patch on one synth channel can unpredictably mute or drop the output level of a *different*, unrelated synth channel. It's intermittent — it depends on oscillator-pool fragmentation at the time (which patches are loaded, how many oscillators each needs, and the order patches were loaded/changed in), so some patch choices trigger it and others don't, on the same setup.

## Root cause

`src/patches.c`, in the memory layout set up during init:

```c
osc_to_voice = (uint8_t *)(memory_patch_oscs + max_num_memory_patches);
voice_to_base_osc = (uint16_t *)(osc_to_voice + AMY_OSCS);
```

`voice_to_base_osc` is allocated **immediately after** `osc_to_voice`, with no padding.

In `patches_load_patch()`'s free-block search (around line 1238):

```c
for(uint16_t i=0;i<AMY_OSCS;i++) {
    uint16_t osc = (osc_start + i) % AMY_OSCS;
    if(AMY_IS_UNSET(osc_to_voice[osc])) {
        good = 1;
        for(uint16_t j=0; j < oscs_per_voice; j++) {
            good = good & (AMY_IS_UNSET(osc_to_voice[osc + j]));   // <-- osc+j not wrapped
        }
        if(good) {
            voice_to_base_osc[voices[v]] = osc;
            for(uint16_t j=0; j < oscs_per_voice; j++) {
                osc_to_voice[osc+j] = voices[v];                   // <-- osc+j not wrapped: OOB write
                schedule_osc_reset(e->time, osc + j, NULL);
            }
            i = AMY_OSCS + 1;
        }
    }
}
```

[...]

PR #967, opened 2026-07-24
Avoid overrunning osc table (#966); default `num_voices` to 1; let drums set their own `oscs_per_voice`.

dpwe:
There was a bug in the code finding blocks of available oscs when allocating voices in `patches_load_patch`; code was reworked.  Testing revealed that loading drum patches involved allocating twice, because the drum patches start by setting their own `num_voices` and `oscs_per_voice`.  Made `num_voices` default to 1 if not specified (and if no existing value) in `patches_load_patch`.  This simplifies loading drum patches to `amy.send(synth=10, patch=384)` etc. Updated documentation.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1230

PR #982, opened 2026-07-26
Fix the 49.7-day millisecond-clock rollover

bwhitman:
`amy_sysclock()` is u32 milliseconds, so it wraps every 2^32 ms = **49.71 days**. Two separate things broke at that instant, and neither is reachable from the audio-rendering test suite because you can't render 50 days of samples.

Came out of a "can an AMYboard stay up for 10 months?" question. As of today: no — it goes half-dead at 49.7 days.

### 1. The sequencer stopped permanently

`sequencer_check_and_fill()` anchored its 64-bit `next_amy_tick_us` accumulator to the 32-bit clock. At the rollover `now_us` collapsed to ~0 while `next_amy_tick_us` stayed at ~4.3e12 µs, so neither the catch-up guard (which only handles falling *behind*) nor the tick loop could fire. It would only recover after another 49.7 days, at which point it wraps again — so, permanently.

Measured against the current code by fast-forwarding `total_blocks`:

```
uptime         sysclock(ms)    ticks/10s
1190.000    h  4284009981            862     <- healthy
1193.000    h  4294809983            862     <- healthy
crossing       4294947276 -> 9934   1730/30s
wrap + 10 s    9934                   0
wrap + 1 h     319456                 0      <- permanently dead
```

Downstream in tulipcc this also takes out `tulip.defer()` and every `tulip.seq_add_callback()`, which ride AMY's sequencer hook — the board still answers live MIDI, but nothing sequences anymore.

**Fix:** add `amy_sysclock64()`, a non-wrapping 64-bit ms clock (`total_blocks` is u32 and ticks once per block → ~219 years of headroom at 44.1 kHz), and anchor every `next_amy_tick_us` site to it. `amy_sysclock()` is now just its low 32 bits, so the public/wire API is unchanged.

### 2. A note could stick on forever

The delta queue is time-sorted on u32 ms with a plain `>=`. A `note_off` scheduled a few ms past the rollover has a tiny `d->time` while its own `note_on` is still near 2^32, so the release sorted *ahead* of the attack, fired first, and left the note droning:

```
sysclock=4294966793: note_on @4294966793, note_off @497 (note_off time wrapped)
  +0.2s sysclock=4294966990 osc1 audible=1
  +2.2s sysclock=1691       osc1 audible=1  <- crossed the wrap
  +32s  sysclock=31645      osc1 audible=1
```

**Fix:** comparisons are now wrap-relative via `AMY_TIME_GEQ` (the kernel's `time_after()` idiom), valid for anything within ~24.8 days of now.

Also dodges the u32 "unset" sentinel: `UINT32_MAX` means "no time set" for an event, and the clock really does land on it for one millisecond every 49.7 days.

### Not a bug (verified)

`amy_global.total_samples` is u32 and wraps every 27.05 h, but envelope math is `total_samples - note_on_clock` in u32, which is modular-correct. A held note sails straight through `4294967040 → 0` with no change in output. The new test pins this down as a control so nobody "fixes" it later.

### Testing

- `make test` — 119/119 pass, all `err=-100.0 dB` (bit-identical output).
- `make check-c-api` / `make godot-api` — no drift.
- New `tests/test_clock_wrap.c` + `make ctest`: plain C tests for things the wav-comparison suite can't reach. Fast-forwards `total_blocks` to the rollover and checks the sequencer keeps ticking, the note stops, scheduling still works afterwards, and the 27-hour wrap stays benign. Reverting either fix above fails it (confirmed — 3 checks go red).

Companion tulipcc PR repins `amy` here and fixes a related `int32_t get_ticks_ms()` that went negative at 24.9 days.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ amy-example
 amy-message
 failed_tests.txt
 tests/tst
+tests/test_clock_wrap
 amy-piano
 src/build/
 *.DS_Store*
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ EMSCRIPTEN_OPTIONS = -s WASM=1 --bind \
 -s ASYNCIFY -s ASYNCIFY_STACK_SIZE=128000
 PYTHON = python3
 
-.PHONY: default all clean amy-module test web deploy-web godot-api c-api check-c-api
+.PHONY: default all clean amy-module test ctest web deploy-web godot-api c-api check-c-api
 
 default: $(TARGET)
 all: default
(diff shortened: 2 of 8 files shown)


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1243

PR #983, opened 2026-07-26
interp_partials: don't render partials that were never allocated

bwhitman:
Three ways `render_partials()` can walk oscs it shouldn't, all reachable when a big partials patch meets a nearly-full osc pool.

Found while investigating a periodic artifact at 8-voice piano polyphony on the AMYboard HW CI bench. **These are real out-of-bounds / uninitialized reads on their own, but they are _not_ confirmed to be that artifact's cause** — see [Scope](#scope-please-read) below. Reviewing this as a standalone hardening change is the right frame.

## The three paths

**1. OOM early-return leaves a stale loop bound.**
`interp_partials_note_on()` returns early when `ensure_osc_allocd()` fails, without ever setting `synth[osc]->last_two[0]` — which `render_partials()` reads as its loop bound. Whatever was there survives: a previous note's partial count, or leftover FM feedback state, since `last_two` is a `SAMPLE` being hijacked to carry an integer. The render then walks oscs this note never set up. Now zeroed on every early-out.

**2. `render_partials()` had no NULL guard.**
`partials_note_on()` deliberately skips partials that fail to allocate (`// On OOM this partial stays silent`), leaving those `synth[]` slots NULL. `render_partials()` then does:

```c
if(synth[o]->role == SYNTH_IS_ALGO_SOURCE) {
```

Both `partials_note_off()` and `interp_partials_note_off()` already guard this (`if (synth[o] == NULL) continue;` / `if (synth[o])`) — `render_partials()` was the odd one out.

**3. The bounds check doesn't exist in firmware.**
The loop lost its `% AMY_OSCS` wrap when it was unrolled for speed, and the `assert()` that replaced it is compiled out of every firmware build (`-DNDEBUG` is in the ESP-IDF flags), so an out-of-range count reads straight past `synth[]`. The count is now clamped to the pool in `render_partials()`, and both `note_on` paths refuse to reserve past the end.

Also: `partials_note_on()` stashed `synth[osc]->preset` rather than the number of partials it actually reserved. It now stores the clamped count.

## Testing

`make test` → **119 tests pass, all `err=-100.0 dB`** (bit-identical to the committed references). Behaviour is unchanged whenever the partials fit and allocate, which is every existing test — all the new code is guards on paths that don't trigger normally.

Also builds clean in tulipcc (macOS desktop target, which compiles AMY from source).

## Scope: please read

The artifact that prompted this — periodic transients at **167.5 ms** with `patch=256, num_voices=8` and 8 low notes held, reproduced twice on the HW CI bench at phase coherence ~0.89, with render load only 0.185 mean / 0.52 max — is **not** explained by these paths.

`patch_oscs[256] = 25`, so 8 voices reserve **200 oscs** against a default `max_oscs = 250`. The OOM path should not trigger there at all. My earlier osc-pool-exhaustion theory was wrong: `_max_partials_for_partials_voice()` returns a per-*preset* constant (24 partials, from `use_this_partial_map` over `num_harmonics[0]`), not a per-note count, so a voice reserves 25 oscs regardless of pitch.

So this PR is hardening for the ≥10-voice / larger-preset case. The 167.5 ms artifact is still open.

### DIFF
--- a/src/interp_partials.c
+++ b/src/interp_partials.c
@@ -43,6 +43,9 @@ const bool use_this_partial_map[MAX_NUM_HARMONICS] = {
 // choose a preset from the .h file
 void partials_note_on(uint16_t osc) {
     int num_partials = synth[osc]->preset;
+    // Never reserve or render past the end of the osc pool (see render_partials).
+    if (osc + 1 + num_partials > AMY_OSCS)
+        num_partials = (osc + 1 < AMY_OSCS) ? (AMY_OSCS - osc - 1) : 0;
     for (int i = 0; i < num_partials; ++i) {
         int o = osc + 1 + i;
         // On OOM this partial stays silent.
@@ -57,8 +60,9 @@ void partials_note_on(uint16_t osc) {
         msynth[o]->logfreq = synth[o]->logfreq_coefs[COEF_CONST] + msynth[osc]->logfreq;
         partial_note_on(o);
     }
-    // Squirrel away num_oscs
-    synth[osc]->last_two[0] = synth[osc]->preset;
+    // Squirrel away num_oscs -- the count we actually reserved, which the pool
+    // clamp above may have reduced below preset.
+    synth[osc]->last_two[0] = num_partials;
 }
 
 void partials_note_off(uint16_t osc) {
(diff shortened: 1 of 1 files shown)


bwhitman:
just fyi this doesn't fix the 8 poly piano click, but it's something to look into anyway

dpwe:
Thanks. Kinda thing that gets on my nerves a bit. The assert is only there to stop the reader worrying that it hasn't been considered. By construction, synth voice allocs don't straddle the end of the osc pool. Also, by convention we don't intend to survive oscs that failed allocation, although rt-rtos has been trying to patch that up. 

But I appreciate you passing it on.

bwhitman:
Closing per Dan — not worth the churn. Nothing here was merged, and tulipcc's temporary pin to this branch is being reverted back to amy main (86c0bd8).

For the record, the three issues are still latent, just not judged worth fixing now: `interp_partials_note_on()` early-returns on OOM without setting `last_two[0]` (which `render_partials()` uses as its loop bound), `render_partials()` has no `synth[o]` NULL guard where both note_off paths do, and its `assert()` bounds check is compiled out under `-DNDEBUG` in every firmware build. All only reachable at >=10 voices / larger presets.

Separately: these were confirmed NOT to be the cause of the 8-voice piano crackle — HW CI showed the 167.5ms artifact completely unchanged with this branch pinned.

PR #987, opened 2026-07-27
oscillators.c: render_lut_256 shaves cycles off sine partial rendering.

dpwe:
By having a special case `render_lut_256` in which the lookup table is assumed to be 256 points with an extra 257th point at the end to avoid having to wrap the index, we're able to bring 6-note speediest-piano from 4068us down to 3412us.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,7 +135,7 @@ speedtest:
 
 speedtest-piano:
 	echo "Compiling LoadTestChord.ino..."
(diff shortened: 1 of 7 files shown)


dpwe:
When flashed onto AMYboard, this version of AMY does 8 simultaneous piano notes with no dropouts.

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1242

PR #995, opened 2026-07-27
src/oscillators: Shaved another op from render_lut_256; use for all sines.

dpwe:
`speedtest-piano` imporves from 3416us to 3258us by moving one bit shift out of loop.   Also, use `render_lut_256` for plain sine and FM unmodulated sine too, since it's faster.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1253

PR #1000, opened 2026-07-28
Add FILTER_NOTCH - RBJ second-order notch filter type

rt-rtos:
This PR is made on the assumption that the notch's absence was
architectural rather than a design choice.

Updated: Added one line JS Notch binding to fix the failing CI test

New SMUL64R precision allows us to meet the  precision demands
of implementing the notch filter.
Its depth is exactly how precisely the zeros land on the unit circle; 
the truncating `SMULR6` multiply left a notch shallow-to-boosting at 
low center frequencies (see the table below), and the quarter-sine
LUT's absolute error shifts a low-f0 center enough for a high-Q notch to miss its own bandwidth.
Both limits have since fallen: #951 gave the kernel an exact multiply under
`AMY_HAS_MUL64`, and the LUT-error half is handled here with a
half-angle identity in the coefficient generator.
With those in place a notch is basically free - the same kernel, the same
`filter_freq`/`resonance` controls, one coefficient generator and a
dispatch branch.

Adds the missing band-reject response alongside LPF/BPF/HPF/LPF24:
`FILTER_NOTCH = 5`, an RBJ second-order notch driven by the same
`filter_freq`/`resonance` controls as the other types. It runs through the
generic split-feedback biquad kernel - the notch's middle feedforward tap
(`b1 = -2 cos w0`) moves with frequency, so the fixed-zeros LPF24/EQ fast
paths don't apply.

One implementation detail worth flagging: the generator recovers `cos(w0)`
via the half-angle identity `1 - 2 sin^2(w0/2)` instead of calling
`cos2pi(w0)` directly. The quarter-sine LUT's absolute error (~2^-15) near
`cos ~= 1` shifts the computed center frequency by several percent at low
cutoffs - enough for a high-Q notch to miss its own bandwidth entirely.
The half-angle form keeps `1 - cos(w0)` at the LUT's relative precision.
Measured on the same LUT path: -2.9 dB rejection at 80 Hz / Q=8 with plain
`cos2pi`, >= 60 dB with the half-angle form.

`amy/constants.py` regenerated (picks up `FILTER_NOTCH=5`).

### Fixed-point: SMUL64R vs SMULR6 (`FILT_MUL_SS`)

The notch runs through the shared biquad kernel, so on fixed-point builds
its multiply is whatever `FILT_MUL_SS` resolves to: exact `SMUL64R` under
`AMY_HAS_MUL64` (#951), truncating `SMULR6` otherwise. That matters more
for a notch than for the other types - rejection depth is set by how
precisely the zeros land on the unit circle, so coefficient/product
truncation shows up directly as lost depth at center.

Bit-exact simulation of the kernel (s8.23, 48 kHz, libm-accurate
coefficients; sine in, steady-state level vs unfiltered baseline):

| Notch center f0 | Q | Probe sine | SMUL64R | SMULR6 | double (ref) |
|---|---|---|---|---|---|
| 80 Hz | 0.7 | at f0 | -84.3 dB | +7.4 dB | -137.7 dB |
| 80 Hz | 0.7 | ½ octave up | -7.0 dB | +4.0 dB | -7.0 dB |
| 80 Hz | 2.0 | at f0 | -59.0 dB | -2.5 dB | -137.5 dB |
| 80 Hz | 2.0 | ½ octave up | -1.8 dB | +8.7 dB | -1.8 dB |
| 80 Hz | 8.0 | at f0 | -39.0 dB | -1.8 dB | -137.5 dB |
| 80 Hz | 8.0 | ½ octave up | -0.1 dB | +17.0 dB | -0.1 dB |
| 500 Hz | 0.7 | at f0 | -93.2 dB | -30.0 dB | -138.0 dB |
| 500 Hz | 0.7 | ½ octave up | -7.1 dB | -6.8 dB | -7.1 dB |
| 500 Hz | 2.0 | at f0 | -81.9 dB | -30.5 dB | -137.8 dB |
| 500 Hz | 2.0 | ½ octave up | -1.8 dB | -1.7 dB | -1.8 dB |
| 500 Hz | 8.0 | at f0 | -72.3 dB | -13.6 dB | -137.8 dB |
| 500 Hz | 8.0 | ½ octave up | -0.1 dB | +0.4 dB | -0.1 dB |
| 2000 Hz | 0.7 | at f0 | -123.8 dB | -58.1 dB | -138.5 dB |
| 2000 Hz | 0.7 | ½ octave up | -6.9 dB | -7.0 dB | -6.9 dB |
| 2000 Hz | 2.0 | at f0 | -125.8 dB | -57.7 dB | -138.1 dB |
| 2000 Hz | 2.0 | ½ octave up | -1.7 dB | -1.7 dB | -1.7 dB |
| 2000 Hz | 8.0 | at f0 | -103.5 dB | -41.2 dB | -138.0 dB |
| 2000 Hz | 8.0 | ½ octave up | -0.1 dB | -0.1 dB | -0.1 dB |

(The double column's ~-138 dB is the harness's own float-input noise
floor, not a filter property.)

Read on the two paths:
- **SMUL64R (exact multiply):** -39 dB worst case (80 Hz / Q=8) improving
  to -126 dB; the residual limit is s8.23 coefficient quantization, not
  the kernel. Off-center response matches double to 0.1 dB. This is the
  behavior on any target defining `AMY_HAS_MUL64` (ESP32-S3 etc.) and on
  float builds.
- **SMULR6 (truncating fallback):** collapses at low f0 - at 80 Hz the
  "notch" not only fails to reject but can BOOST, up to +17 dB resonant
  gain at Q=8 half an octave above center. Usable-ish above ~1 kHz,
  wrong below.

### Float-path accuracy (half-angle fix)

Host-sim, float build, steady-state vs FILTER_NONE baseline, probe sine
at f0:

| Notch center f0 | Q=0.7 at f0 | Q=2 at f0 | Q=8 at f0 |
|---|---|---|---|
| 80 Hz | -66.7 dB | -60.1 dB | <-240 dB |
| 500 Hz | <-240 dB | <-240 dB | -59.2 dB |
| 2000 Hz | <-240 dB | <-240 dB | <-240 dB |

Half an octave off-center: -7 dB (Q=0.7) / -1.8 dB (Q=2) / -0.1 dB
(Q=8), independent of f0.

(<-240 dB = below the int16 quantization floor.) Before the half-angle
fix the same grid read -21.0 / -12.2 / -2.9 dB at 80 Hz for Q 0.7/2/8 -
the motivating failure.

### Other verification
- Implented and groove-tested on device in my S3 project
- All 119 reference-test results are bit-identical to main on the same
  host/toolchain (the change adds an unused define and an uncalled function
  until an event selects type 5).
- Functional smoke via the Python module: 500 Hz sine through
  `filter_type=FILTER_NOTCH`, Q=8 - -32 dB at center, -0.13 dB with the
  notch parked at 2 kHz.
- ESP32-S3 codegen (esp-gcc 15.2, -O2): generator is FP64-free
  (sin_lut + __divsf3 only), matching the other generators post-470a6c0.

Happy to add a reference test (e.g. swept sine through a fixed notch) if
that's the preferred shape for new filter types.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1258

dpwe:
I was a bit disappointed by the sound - it's quite a subtle effect, don't you think?

rt-rtos:
I agree. I have not yet built an intuition for what it "should" sound like; I was aware that it's considered the least broadly useful of the fundamental filter types - so I didn´t have the highest expectations.

I was pondering potential improvements earlier in the day. The one thing that came to mind as a possibility was a cascaded notch filter - a staggered pair offset either side with Q setting the spacing, would have more character, since you would get an attenuated band and center bump instead of a point.

This would require a new kernel though, and would be hard to justify given the low impact of this filter type.
Probably more work and less DSP bang for buck than something like a cascaded allpass phaser with feedback and LFO.

PR #1002, opened 2026-07-29
Tidy wire formatting: floats always trunc, coefs accept tuples

bwhitman:
Follow-up to review discussion on shorepine/tulipcc#1256 (the C `amy.message()` fast path), which reproduced three asymmetries in the Python builders. Per @dpwe: tidy them instead.

- **Coefs accept tuples**: `parse_ctrl_coefs` asserted on tuples while `parse_list_or_comma_string` joined them; both now accept `(list, tuple)`.
- **Floats always trunc**: a scalar coef went through `trunc()` while list/coef elements went through `str()`, so `freq=440.0` made `f440` but `freq=[440.0]` made `f440.0` — and `str(float)` is repr, which differs between CPython and MicroPython and can emit exponent notation (`1e-07`) the wire parser has no use for. Every float element now truncs to the same canonical form via one shared `elem_to_str()`.
- **Unchanged**: `None` → `''`, non-floats → `str()` (so filenames and `'%'`-tokens in lists are untouched), and the `'%'`-token passthrough stays.

Wire strings only get shorter (`bp0=[0,1.0,500,0.5]` → `A0,1,500,0.5`), visible in the regenerated `patches.h` / `patches.generated.js` committed here.

`make test`: 119 tests pass. tulipcc's C fast path will mirror this in shorepine/tulipcc#1256, which will pin to this change.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1257

PR #1017, opened 2026-07-31
sequencer: walk the slots that are scheduled, not 0..highest_tag

bwhitman:
*(Reworked on top of the wire-string sequencer + anonymous `ticks=` pool now on main; also absorbs the regression coverage from #1016, which is closed — its fix already landed.)*

## The problem

`sequencer_process_tick()` sweeps the whole table up to a high-water mark:

```c
for (int32_t tag = 0; tag <= highest_tag; ++tag) {
```

`highest_tag` is set when a slot is used and **never lowered when it's cleared or fires**. The anonymous pool makes this the common case rather than a corner: anonymous `ticks=` one-shots — now the primary way to schedule anything — are allocated round-robin at indices *past* `max_sequences`, so a burst of scheduled events pins the mark at the very end of the table permanently. With the default 256 tags that's a 512-entry sweep on every tick for the rest of the session, and raising `max_sequencer_tags` makes it proportionally worse.

## The change

The occupied slots — user tags and anonymous entries alike — are threaded through the table itself as an **ascending list**: one `int32_t next_active` per entry plus a `first_active` head. Per-tick cost now tracks what's actually scheduled; an anonymous one-shot that has fired leaves nothing behind.

## Why threaded, why ascending, and the locking story

**The table has to stay indexable** — add and clear both reach a tag directly and want O(1).

**Ascending is not tidiness**: two sequences that hit on the same tick play in visit order, which decides who wins if they touch the same parameter. That order was slot order under the sweep, and the sorted list keeps it slot order — an insertion-ordered list would make a pattern sound different after an edit.

**Locking**: link mutations all happen under the amy lock (`sequencer_add_wire` takes it; the tick loop's one-shot delete takes it; `sequencer_reset` is called with it held from `play_delta`). The tick *walk* stays lockless, which is safe because the links are indices into a fixed array, not pointers: a splice publishes in one aligned 32-bit store, and every stored link is greater than the slot holding it, so a stale read can skip or revisit an entry for one tick but cannot cycle, hang, or leave the array. A list of malloc'd nodes would be a different hazard class — a torn `next` pointer walks the render thread into freed memory.

## Also fixed along the way

Porting #1016's bounds test flushed out a live bug: `sprint_event` printed `ticks` values as **signed**, so a C-API tag past INT32_MAX serialized with a `-`, the unsigned list parser stopped at it, and the (invalid, should-be-rejected) 3-value ticks silently became a 2-value **anonymous** entry — stored and fired instead of refused. `ticks` now prints unsigned (`_EPRINT_U_SEQ`) and the out-of-range tag is rejected by `sequencer_add_wire`'s existing unsigned bounds check.

## Testing (`make ctest`)

- **`tests/test_sequencer_active.c`** — tags added out of order all fire and clear removes only one; anonymous one-shots fire once and leave `first_active == -1`; and the headline invariant: a lone sequence at tag 4095 costs the same as one at tag 0 (measured at high tempo so the scan dominates; **15x apart** under the old sweep, ~1x now).
- **`tests/test_sequencer_bounds.c`** (from #1016, ported to `ticks=`) — tag 0 and max−1 accepted; max, max+1, far-out and past-INT32_MAX tags all rejected; and an out-of-range user tag can't clobber the anonymous entry sitting right past the user range.

`make test`: **122 WAV tests pass bit-exact** (`err=-100.0 dB` on all `TestSequencer*`).

## Cost

4 bytes per table entry. Insert is an O(active) walk under the lock on the caller's thread — microseconds at realistic counts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,7 +104,7 @@ amy-message: $(OBJECTS) src/amy-message.o
 
 # Plain C tests for things the audio-rendering suite can't reach -- e.g. clock
 # rollovers 50 days out, which you can only hit by fast-forwarding the counters.
-CTESTS = tests/test_clock_wrap
+CTESTS = tests/test_clock_wrap tests/test_sequencer_active tests/test_sequencer_bounds
 
 # Static pattern rules, so these win over the generic %.o: %.c above (which
 # would compile without -Isrc and fail to find amy.h).
--- a/src/patches.c
+++ b/src/patches.c
@@ -222,6 +222,28 @@ void snprintfloat3dp(char *s, size_t max_len, float val) {
         } \
     }     \
 }
+// As _EPRINT_I_SEQ but unsigned. ticks needs this: its values are uint32_t,
+// and printing one past INT32_MAX as a negative number makes the unsigned
+// list parser on the other end stop at the '-' -- for a 3-value ticks that
+// silently turned an (invalid, should-be-rejected) tag into a 2-value
+// anonymous entry.
+#define _EPRINT_U_SEQ(FIELD, NAME, LEN, WIRECODE) {      \
+    int last_set = -1; \
+    for (int i = 0; i < LEN; ++i) {    \
+        if (AMY_IS_SET(e->FIELD[i])) last_set = i; \
+    }                                              \
+    if (last_set >= 0) { \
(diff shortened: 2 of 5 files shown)


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1270

PR #1019, opened 2026-08-01
Replace `time=` and `sequence=` with `ticks=`

dpwe:
**DO NOT MERGE** — for review only right now.

## Summary
- Removes the millisecond `time=` scheduling kwarg (wire `t` command, the C-API's far-future `e->time` deferral, and the ms-based `timed_wire_*` store) entirely.
- Renames `sequence=` to `ticks=`, and adds two new forms alongside the existing `tick,period,tag`:
  - `ticks="tick"` — one-off, anonymous
  - `ticks="tick,period"` — repeating, anonymous
  Anonymous entries live in a small reserved pool separate from the public tag range, so they can never collide with or be cancelled via a user-supplied tag.
- Simplifies `amy_add_message()` to a straight-line function now that transfer-status and leading-schedule checks are call-level, not per-segment.
- Migrates `amy/test.py` (via a new `amy_send_at()` helper that renders forward to reproduce the old timing bit-exactly), plus `amy/examples.py`, `amy/piano.py`, `amy/xanadu.py`, `experiments/tulip_piano.py`, `experiments/piano_examples.py`, and `docs/tutorial.html` off the removed kwargs.

## Test plan
- [x] `make test` — all 120 WAV tests pass (one reference, `TestVoiceStealDecay`, regenerated — see commit message for why)
- [x] `make ctest` — clock-wrap suite passes
- [x] Manual smoke tests of the live-demo paths (`amy.examples`, `amy.piano`) against the offline engine

## Known follow-ups (not in this PR)
- `tulip/shared/py/_boot.py` in the outer `tulipcc` repo has a fast-path hook (`insert_time`) that references the removed `time=` mechanism — needs updating once this is pinned into `tulipcc`, per the submodule policy.
- A few pre-existing, unrelated bugs surfaced while touching these files (missing imports in `amy/examples.py`'s `play_example_patches`/`play_baked_patches`; a nonexistent function call in `experiments/piano_examples.py`'s `init_piano_voices`; a stale `store_patch` keyword in `experiments/tulip_piano.py`) — left as-is, out of scope here.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1264

PR #1020, opened 2026-08-01
Add FILTER_PHASER: 6-stage allpass phaser filter type

rt-rtos:
## Spectrogram
Screenshot 2026-08-01 190204

*Triangle Sweep LFO with Chained Osc LFO Wobble - Low and High Q*

## What

A new per-oscillator filter type, `FILTER_PHASER = 6`: six cascaded
first-order allpass stages with global feedback and a fixed 50/50 dry/wet
mix - the classic 3-notch phaser. The notch centers track `filter_freq`
like every other filter type, so all existing modulation routes (envelopes,
mod oscillators, pitch bend) sweep it with no new code, and feedback
("regeneration") is mapped from the existing `resonance` parameter. No new
parameters, no state growth.

```mermaid
flowchart LR
    X["x[n]"] --> SUM(("+"))
    SUM --> AP["AP1 .. AP6<br/>first-order allpass x6<br/>shared coefficient a"]
    AP --> YC["y[n]"]
    YC -- "fb * z^-1" --> SUM
    X --> MIX(("0.5 * (x + y)"))
    YC --> MIX
    MIX --> OUT["out[n]"]
```

Each stage is the one-multiply transposed form of
`H(z) = (a + z^-1) / (1 + a z^-1)`:

```
y0 = a*x0 + w[k];   w[k] = x0 - a*y0
```

with `a = (tan(pi*f) - 1) / (tan(pi*f) + 1)` computed once per block from
`filter_freq` (each stage contributes -90 degrees at `f`, putting the middle
notch at the cutoff), and
`fb = clamp(0.85 * (Q - 0.51) / (8.0 - 0.51), 0, 0.85)` from `resonance`.

## Why

AMY currently has no phaser-style effect, and a swept multi-notch response
is not reachable from the existing filter types. This implementation is
cheap: 13 multiplies per sample (6 stages x 2, plus feedback) against 5 for
the existing 12 dB biquad types - about 2x the single-biquad render cost as
measured on the ESP32-S3 fixed-point build - plus one `tan` (via the
existing `sin2pi`/`cos2pi` LUT helpers) and two float divides per block.

Offering it upstream after implementing it for my fork. 

## Implementation notes

- **Not a biquad.** It takes a dedicated branch in `filter_process()` ahead
  of the coefficient generators (it produces no `coeffs[5]`), implemented as
  `allpass1_chain()` (the reusable stage-chain primitive) wrapped by
  `dsps_phaser_f32_ansi()`.
- **State fits the existing `filter_delay[2 * FILT_NUM_DELAYS]`:** six stage
  memories in words 0-5, the feedback sample in word 6. `reset_filter()` and
  the silence early-out (`scan_max` over all 8 words) cover it unchanged.
- **No block-floating-point wrapper.** The chain is unity-gain at every
  frequency and the feedback loop gain is `fb < 1`, so the filter is
  unconditionally stable and the worst-case resonant gain `1/(1-fb) <= 6.7`
  stays well inside fixed-point headroom. There is also no LPF-style DC
  accumulation.
- **Notch depth is structural.** The same `a` feeds both multiplies of each
  stage, so every stage is exactly allpass for any representable `|a| < 1`;
  coefficient quantization moves the notch centers slightly but can never
  fill the notches in. Consequence for non-`AMY_HAS_MUL64` targets: the
  `SMULR6` fallback only coarsens notch-center placement at low center
  frequencies (a few cents) - depth is unaffected, unlike the known shallow-
  notch behavior of pole-zero filters under `SMULR6`.

## Verification (host, fixed-point build, 44.1 kHz)

Sine tones driven through the full `amy_add_event` -> render path, measured
against the analytic transfer function
`H = 0.5 * (1 + Hc / (1 - fb*Hc*z^-1))`, `Hc = H1^6`.

**fb = 0 (Q at the floor): null depth and passband flatness**

| filter_freq | analytic nulls (Hz) | measured depth (dB) | passband vs analytic |
|---|---|---|---|
| 200 Hz | 53.5 / 200.1 / 745.8 | -52 / -54 / -65 | within 0.08 dB |
| 1 kHz | 268.5 / 1000.0 / 3653.7 | floor / -71 / floor | within 0.04 dB |
| 4 kHz | 1099.7 / 3999.9 / 11649.6 | floor / floor / floor | within 0.05 dB |

("floor" = output below the int16 measurement floor; measured nulls sit at
the analytic frequencies.)

**Feedback from resonance (filter_freq = 1 kHz), 14-point log grid**

| Q | fb | max \|measured - analytic\| (nulls excluded) | peak gain measured vs analytic |
|---|---|---|---|
| 2.0 | 0.169 | 0.28 dB | +0.81 dB vs +0.81 dB |
| 8.0 | 0.850 | 0.11 dB | +10.75 dB vs +10.75 dB |

Stable at the fb cap; Q at the floor is bit-identical to fb = 0.

**Codegen (ESP32-S3, gcc -O2 + LTO):** the chain fully inlines into
`filter_process`; the 6-stage inner loop compiles to an Xtensa hardware
zero-overhead loop (12-instruction body, no calls).

## Docs

`docs/api.md`: `G` (filter type) range extended to 0-6, "6 = phaser".

## Tests

Happy to add an `amy/test.py` case with a reference WAV (along the lines of
`TestNotchFilter`) if you want one in this PR - say the word.


dpwe:
Looks awesome!!

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1265

Issue #1028, opened 2026-08-02
Remove amy.inject_midi; use amy.inject_midi_bytes


dpwe:
`amy.inject_midi` accepts a time arg i think purely to support tests. Tests don't need that any more, and I'm trying to get rid of time plumbing in midi events.

PR #1040, opened 2026-08-02
midi: drop the dead time argument, and inject_midi with it

dpwe:
## The `time` argument was already dead

In `amy_midi.c`, `time` is declared `AMY_UNSET_VALUE(time)` (lines 321, 395) and **never reassigned** — so every real MIDI path (USB, gadget, UART, mac, sysex) has always passed it unset. `amy_received_control_change()` didn't even read its copy. Nothing outside `src/` called these at all. The one caller that ever supplied a real value was pyamy's `inject_midi()`, i.e. the test suite.

That reflects the truth of the thing: **a live MIDI message has no time of its own — it plays when it arrives.** So the argument is gone from `amy_event_midi_message_received()` and the five `amy_received_*` handlers.

## `inject_midi` goes with it

It was the weaker of the two Python entry points:

- It hardcoded `len=3`, so it could not send a 2-byte message at all — `test.py` already used `inject_midi_bytes([0xC0, 5])` for program change.
- It handed a pre-formed message straight past the byte-stream parser.

`inject_midi_bytes()` runs the real `convert_midi_bytes_to_messages()`, so the tests now exercise running status and real-time interleaving the way actual MIDI input does. Net gain in coverage.

The ~16 test call sites that passed an absolute millisecond time move to a new `amy_inject_midi_at()`, the MIDI twin of the existing `amy_send_at()`: render up to the moment, then inject.

## What deliberately did NOT change

`midi_message_handler_to_queue()` **keeps** its `time` parameter. It lives in `midi_mappings.c`, and `patches.c:1081` drives it with a real `e->time` for the `wave=AMY_MIDI` osc — that's how a tick-scheduled event emits MIDI on the beat. `amy_event_midi_message_received()` now passes it an explicit unset instead of a laundered one.

## Drive-by fixes

- `pyamy.c` had three bare `return Py_None;` (missing INCREF — each call over-decremented `None`'s refcount). Now `Py_RETURN_NONE`.
- The Daisy port's `midi_polyphony()` demo was staggering notes 1000ms apart with the argument that never did anything; it now says so instead of pretending.

## Testing

`make test`: **122 tests pass**, and **all 116 waveform comparisons are bit-exact** (`err=-100.0 dB`) — no reference audio needed regenerating, despite the tests now routing through the byte-stream parser and the render-forward clock.

`make check-c-api`: no generated-API drift. Clean `make` with no new warnings.

### DIFF
--- a/amy/__init__.py
+++ b/amy/__init__.py
@@ -397,16 +397,11 @@ def restart(default_synths=0):
     _amy.stop()
     _amy.start(default_synths)
 
-def inject_midi(a, b, c, d=None):
-    if d is None:
-        _amy.inject_midi(a, b, c)
-    else:
-        _amy.inject_midi(a, b, c, d)
-
 def inject_midi_bytes(data, usb=0):
     # Feed a raw MIDI byte stream (list/tuple/bytes of ints) through AMY's
-    # byte-stream parser, exercising running status and real-time interleaving --
-    # unlike inject_midi(), which injects a single pre-formed message.
+    # byte-stream parser, exercising running status and real-time interleaving
+    # just as real MIDI input does. There is no scheduling argument: live MIDI
+    # has no time of its own, it plays when it arrives.
     _amy.inject_midi_bytes(data, usb)
 
 def unload_sample(preset=0):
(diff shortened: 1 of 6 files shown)


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1273

dpwe:
fixed #1028

PR #1045, opened 2026-08-03
Route transfer chunks through the sysex path on every backend

bwhitman:
## The bug

A user hit this loading a wav off SD in an AMYboard sketch:

```
File "/user/current/sketch.py", line 11, in <module>
File "amy/__init__.py", line 543, in load_sample
File "amy/__init__.py", line 464, in _send_transfer_chunk
File "_boot.py", line 111, in <lambda>
OSError: [Errno 19] ENODEV
```

The wav read fine — the failure is in *delivering* the sample to AMY.

`_send_transfer_chunk()` fell back to `_amy.send_wire_from_sysex()`, the CPython-only `c_amy` binding. On linked MicroPython (Tulip/AMYboard hardware) and on web, `_amy` is `None`, and the platform's default transport was installed as `amy.override_send` (tulipcc `_boot.py:111`, both `spss.js`), so every base64 chunk took the `override_send` branch into `amy_add_message()` with `sysex=false`.

`amy_add_message_with_sysex_flag()` only routes to `parse_transfer_message()` when that flag is set. So chunks were parsed as ordinary wire commands.

That is worse than a silent no-op: base64 contains `z`, AMY's transfer-layer prefix (`parse.c:793`), so a chunk eventually parses as `zF`/`zD`/`zT` and hands a garbage filename to the platform file hooks. On AMYboard those go `mp_fopen_hook` → `mp_vfs_open` with no `nlr_push`, so the OSError propagates out of whatever Python frame happened to be running — the `amy.send` lambda. `ENODEV` rather than `ENOENT` is the tell: AMYboard mounts `/sys`, `/user`, `/sd` and no `/`, so MicroPython reports "no mount matches this path" for a base64-shaped filename. AMY is also left wedged with `transfer_flag` set, eating subsequent real sysex.

## The fix

- `send_wire_from_sysex` gains `mp` and `web` platforms in the codegen table, so `tulip.amy_send_wire_from_sysex` and the JS/`AMY_C_API_PY_INSTALL` binding now exist (`make c-api` outputs included).
- `_send_transfer_chunk()` uses the resolved `_send_wire_from_sysex` backend, which reaches the local AMY on CPython, linked MicroPython and web alike.
- `override_send` keeps its documented meaning — the user redirected AMY to *another* board over MIDI sysex or i2c — and chunks still follow it there, getting marked by the far end's transport.

Callers must stop installing the platform default into `override_send` for this to take effect. That is redundant with the generated dispatch (`_send_wire` already resolves to `tulip.amy_send` / the JS binding), and shorepine/tulipcc#1280 drops it from `_boot.py` and both `spss.js`.

Also refreshes `TestLoadSample`'s docstring, which still described the pre-sysex-flag `amy_parsing_from_sysex` design.

## Testing

- `make test`: 124 tests pass.
- `TestLoadSample` covers the CPython route end to end (unmarked chunks → preset never loads → silence). The MicroPython/hardware route this fixes is structurally out of reach for the CPython suite; verified there via tulipcc.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1281

PR #1047, opened 2026-08-04
Make the bus count a runtime config option (amy_config.max_buses)

bwhitman:
## What

`AMY_NUM_BUSES` was a hard-coded `4` in `amy.h` — not even `#ifndef`-guarded, so changing the bus count meant editing the header. It's now **`amy_config.max_buses`**, with **no compile-time ceiling at all**. Default is still 4, so nothing changes unless you ask for it:

```c
c.max_buses = 300;
amy_start(c);
```

Also `amy.live(max_buses=300)` and a `max_buses` property on the Godot `AmySynth` node.

## What was in the way

**The params enum.** Volume was the one bus-directed parameter that encoded its bus *in the param id* — `VOLUME_BASE + bus`, one id per bus — so the bus count was capped by how many ids fit before `MODE` (99), about 33. Every other bus-directed param (`EQ_*`, `ECHO_*`, `CHORUS_*`, `REVERB_*`) already names its bus in `delta.osc`, a `uint16_t`. Volume now does the same: one `VOLUME` id, one delta per bus, bus in `osc`. The wall is gone and ids 66..98 are free. `V1,0.5` still sets consecutive buses — each entry is just its own delta now.

**Fixed-width tables**, which now allocate from `max_buses`: `global_state.bus`/`volume`, `fbl` and `per_osc_fb` (per-core arrays of bus pointers), and the per-block `volume_scale`, which was a stack array sized by the ceiling and is now allocated once. `bus_gain` went with them — it was declared and never read.

**`uint8_t` bus fields** in `synthinfo`, `mod_synthinfo`, `amy_event`, `instrument_info`, `config.max_buses`, `highest_bus` and every `config_*`/filter signature, which would have silently truncated bus 300 to 44. All widened to `uint16_t`, including the public `amy_external_bus_postprocess_hook` — a recompile for anyone implementing it.

### The one constant that stays

`amy_event` keeps a fixed-length volume vector, now `AMY_MAX_VOLUME_LIST` (8) rather than a bus count. It bounds **how many buses a single `V` message can set**, the way `MAX_ALGO_OPS` bounds a message's operator list — not how many buses exist. Any bus is reachable with `bus=n, volume=x` regardless. `amy_event` has to stay a value type (it's returned by value, copied, and stored), so a variable-length member would mean a pointer and an ownership story for something that buys nothing.

## Bus numbers were never range-checked

Pre-existing, and this change had to fix it rather than move it. `parse.c` does `e->bus = atoi(arg)` and the result went straight into `synth[osc]->bus` and `instrument->bus`, both of which subscript `amy_global.bus[]` and `fbl[]`. On the old 4-bus build, `y9` wrote off the end of both.

`amy_validate_bus()` now guards the four places a bus is stored — event→delta (both paths), delta→synth, and both instrument paths — falling back to bus 0 with a warning. Also fixes a `dump_state` line that printed `volume[0..3]` unconditionally, which reads past the array on a build with fewer buses.

## Cost

Per bus: ~3.5 KB of always-allocated mix buffers (~6.5 KB dual-core), plus whatever its FX allocate once switched on (echo + reverb is a few hundred KB). Zero render cost until a bus is used — the per-block loops were already bounded by `highest_bus`.

## Testing

- `tests/test_bus_config.c` (`make ctest`): 300 buses allocated and addressable; the volume vector lands on the right buses from a non-zero start and stops at the last one; bus range enforced over the C API, the wire protocol and both instrument paths; 64 buses rendering reverb + chorus + echo.
- All 124 WAV regression tests **bit-identical**, including the two using `volume='2,0.5'` — the direct check that the VOLUME rework preserves behavior.
- Warning-clean rebuild; `AMY_MAX_VOLUME_LIST` = 1, 2, 16 all build clean.
- `make check-c-api` clean; `amy/constants.py` and `src/amy_api.generated.js` regenerated.

Godot needed the usual four-file lockstep edit (`amy.gd`, `godot/src/*.{h,cpp}`, `setup_godot.sh`); the generated region of `amy.gd` is untouched. `docs/amy.*` is left alone — release automation rebuilds it.


dpwe:
If we go with this, I think we have to retire the "volume is a vector" syntax.  So volume in an event goes back to being a scalar, and the only way to set the volume of a bus other than 0 is with `bus=N, volume=X`.

The big problem with allowing a large number of buses is that each bus gets its own `fbl` and `per_osc_fb` of `AMY_BLOCK` 32-bit samples (so 2KB per bus if my math is correct).  These are in high-speed RAM (since they are where all the oscillator calculations read and write to) which is a scarce resource, and I think we're already running out of it?  (That's why rt-rtos has added all the `malloc` checks).  So even if the compute only scales with the number of buses actually used, the RAM is immediately taken by `MAX_BUSES` and setting it to something large will not go well, I expect.

Have you tried running with lots of buses on ESP32 hardware?

dpwe:
Actually it seems that `c.ram_caps_fbl` is `MALLOC_CAPS_DEFAULT` which is just generic RAM so I guess we can use as much as we want.  The `fbl` buffers are the targets of all the many osc per-sample passes, so avoiding wait-states will significantly impact overall performance; I'm not clear where we're at on that.

dpwe:
We should discuss this further.  My original bus design started from the premise there would be a small number of buses, so I wasn't careful about minimizing the resources of each bus.  I'm not sure there's an alternative, but if you're thinking about having as many buses as oscillators, we might want to rethink the design.

src/amy.h
```diff
@@ -118,10 +118,28 @@ extern void amy_set_gamma9001_pcm(const int16_t * data);
...
+// actually runs is a runtime setting, amy_config.max_buses (AMY_NUM_BUSES
+// below), so a host can trade RAM for buses without rebuilding.
+//
+// There is no compile-time ceiling.  Everything indexed by bus is allocated
+// from max_buses at amy_start, and a bus-directed parameter carries its bus
+// in the delta's uint16_t osc field (see amy_event_to_deltas_queue), so the
+// only limits are RAM and the 65535 that field holds.
+#define AMY_DEFAULT_NUM_BUSES 4
+#define AMY_NUM_BUSES (amy_global.config.max_buses)
+#define AMY_MAX_POSSIBLE_BUSES 65535  // What a delta's uint16_t osc can address.
```
dpwe: I think it's confusing to reference this number of buses, since we will run out of RAM long before reaching this number. I would rather not define this constant at all, and not report any maximum in the range check that is the sole place it is used.

src/amy.h
```diff
@@ -860,7 +886,7 @@ typedef struct global_state {
     amy_config_t config;
     uint8_t running;
     uint8_t i2s_is_in_background;  // Flag not to handle I2S in amy_update.
-    float volume[AMY_NUM_BUSES];  // Volume controls mix of buses into final output.
+    float *volume;  // Per-bus mix into the final output; max_buses entries.
```
dpwe, line 878: AMY_MAX_VOLUME_LIST is an unhappy compromise.  I would prefer we return to making `volume` a scalar, and require specifying `bus` to address `volume` for non-default buses.

src/amy.h
```diff
@@ -118,10 +118,28 @@ extern void amy_set_gamma9001_pcm(const int16_t * data);
...
+// Each bus has separate FX (EQ, chorus, reverb, echo).  How many buses AMY
+// actually runs is a runtime setting, amy_config.max_buses (AMY_NUM_BUSES
+// below), so a host can trade RAM for buses without rebuilding.
+//
+// There is no compile-time ceiling.  Everything indexed by bus is allocated
+// from max_buses at amy_start, and a bus-directed parameter carries its bus
+// in the delta's uint16_t osc field (see amy_event_to_deltas_queue), so the
+// only limits are RAM and the 65535 that field holds.
+#define AMY_DEFAULT_NUM_BUSES 4
+#define AMY_NUM_BUSES (amy_global.config.max_buses)
```
dpwe: I don't like this idiom where something that looks like a compile-time constant turns out to be a variable.  Can you just replace all the instances of `AMY_NUM_BUSES` in the code with `amy_global.config.max_buses`.

dpwe (commented):
Please consider making volume a scalar (so removing AMY_MAX_VOLUME_LIST), removing AMY_MAX_POSSIBLE_BUSES, and using amy_global.config.num_buses in place of AMY_NUM_BUSES throughout.

bwhitman:
Made those requested changes: `volume` is a scalar again.  The vector syntax ("V0.5,0.25" setting consecutive buses) is retired along with AMY_MAX_VOLUME_LIST; the only way to set a non-default bus's mixdown volume is `bus=N, volume=X` (`yNVX` on the wire). AMY_MAX_POSSIBLE_BUSES is gone. The AMY_NUM_BUSES macro is gone

I want to use about 12 buses max on the thing i'm working on now. That's more than 4, which is what started this whole thing. I don't think anyone should have a bus per osc, no. I don't have any nice comparison or metaphor for how many we need to do things people may want, but i think 4 was too little and 32 seems like too much. Just vibes

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1294

PR #1050, opened 2026-08-04
PCM: raise index fractional bits 8 -> 12 - fixes audible detune on pitched-down samples

rt-rtos:
## Summary

The PCM phase accumulator splits its 31 bits as 23 index / 8 fraction. Eight
fractional bits mean the playback step quantizes to `ratio * 256` LSBs and
the interpolation coefficient has only 256 levels. That quantization is
audible as a constant pitch offset that grows as the sample is pitched down.

Measured with a 22050 Hz sample (the TR-808 Bass Drum 1 ROM preset) on a
48 kHz engine, comparing rendered pitch against intent:

| played at | error, 8 frac bits | error, 12 frac bits |
|---|---|---|
| native pitch (midinote 60) | -8.9 cents | <= 2 cents |
| one octave down | -18 cents | <= 2 cents |
| three octaves down (drum floor) | -84.5 cents | <= 2 cents |

Worse than the absolute offset: at the bottom of the range the
*representable-pitch grid* is 90-120 cents coarse, so adjacent semitones
detune by wildly different amounts and pitched bass lines play out of tune
with themselves. (Octave ratios are exactly representable, which hides the
bug in octave-based tests.)

## Change

Split the phase 19.12 instead of 23.8: 16x finer step and interpolation
resolution. The index still covers 524287 frames (~10.9 s at 48 kHz), far
above every baked-in preset. `pcm_load()` now rejects lengths that would
wrap the smaller index range up front instead of glitching silently during
playback.

Zero render cost: the shift amounts are compile-time constants, the
per-sample instruction sequence is unchanged.

## Notes for review

- Noise floor is unchanged (the interp-coefficient zipper at 8 bits was
  already ~-90 dB on bass content; the audible defect is purely the pitch
  quantization).
- `trigger_phase` API interaction: a caller that sets a PCM start frame via
  `trigger_phase` computes it as `start_frame / 2^PCM_INDEX_BITS`, so the
  scaling of that value shifts 16x with this change. Grep of this repo found
  no first-party use that bakes the old constant, but external callers that
  hardcoded `2^23` would need the header constant instead.
- Verified by host-rendering the real ROM preset through the actual render
  path at device rates and measuring zero-crossing pitch against an
  exact-arithmetic reference; steady-state SNR table identical before/after
  except the corrected pitch.

### DIFF
--- a/src/pcm.c
+++ b/src/pcm.c
@@ -143,8 +143,12 @@ void pcm_deinit() {
     pcm_unload_all_presets();
 }
 
-// How many bits used for fractional part of PCM table index.
-#define PCM_INDEX_FRAC_BITS 8
+// How many bits used for fractional part of PCM table index. With 8 bits the
+// step quantizes to ratio * 256 LSBs - an audible pitch offset that grows as
+// a sample is shifted down (up to ~-85 cents three octaves below native).
+// 12 bits measures within 2 cents; 19 index bits still cover 524287 frames
+// (~10.9 s at 48 kHz), enforced in pcm_load().
+#define PCM_INDEX_FRAC_BITS 12
 // The number of bits used to hold the table index.
 #define PCM_INDEX_BITS (31 - PCM_INDEX_FRAC_BITS)
 
@@ -240,7 +244,7 @@ void pcm_note_on(uint16_t osc) {
             // note-on (start_frame / 2^PCM_INDEX_BITS).
             synth[osc]->phase = F2P(synth[osc]->trigger_phase);
         } else {
-            synth[osc]->phase = 0; // s16.15 index into the table; as if a PHASOR into a 16 bit sample table.
+            synth[osc]->phase = 0; // PCM_INDEX_BITS.PCM_INDEX_FRAC_BITS fixed-point index into the table.
         }
         // Copy the looping mode from the wave mode field.  Can be updated on note_off.
         msynth[osc]->state = synth[osc]->mode;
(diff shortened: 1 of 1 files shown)


dpwe:
I don't think 10.9 sec is acceptable as an upper limit on PCM sample length, but I take the point that 8 bits is too few for the sample step.  I am thinking about a different approach to calculating the PCM sample index.

rt-rtos:
I should point out that I stumbled upon this while hunting for the root-cause of a clipping problem that turned out to be a correctness issue in tinyusb + espressif_tinyusb UAC flow control and FIFO buffer size. 

This was discovered not because it is a high-visibility issue that caused problems - but because I was investigating any possible suspect. I don´t want to give the impression that it´s a bigger problem than it is. 

The same is true for the declicker, also found while root-causing something else.

Want me to close one or both of these?

dpwe:
I'm glad to hear this isn't a current major problem for you!

I think the observation that pitch for PCM sounds is quantized at 8 bits is definitely an issue, given that the quantization of computed waveforms is meant to be more like 31 bits.  So I appreciate the issue being brought up.

And the other issue, hard clicks when restarting PCM, is definitely a nasty problem e.g. for the "Blue Monday" scenario.  I'm not super excited about the compensator (although it's a clever design); I'd rather have an actual short fade-out, but that's going to be a lot harder to implement.

dpwe:
The effective PCM frequency resolution was increased to 16 fractional bits in #1064 which should have resolved the issue addressed in this PR (without sacrificing the maximum length of addressable PCM clips, which remains at 8MSa / 190sec @ 44kHz).

rt-rtos:
This is a more elegant solution - widening only in-block, not spending index range, lossless foldback.
I've adopted it on my end and re-ran the harness from this PR: measured detune is now under half a cent across the whole pitched-down range. Thanks for picking this up.

I closed #1051. If you land on the fade-out shape you want, I'll happily take a shot at building it given some steering.

PR #1054, opened 2026-08-07
Eliminate step

dpwe:
`step` and `substep` were archaic fields in `synth[osc]`.  The KS osc still used `step`, but I was able to switch it over to `phase` instead, which I think is the natural place for what it was using. 

I also fixed the initialization `ks_note_on` to use the actual anticipated note frequency which allowed it to correctly pre-fill the KS buffer with noise.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1285

PR #1064, opened 2026-08-08
src/pcm.c: Increase within-block step advance resolution to 16 frac bits, fix end check.

### LINKED ISSUE #1007: PCM_LOOP_FOREVER stops unexpectedly
`PCM_LOOP_FOREVER` used as in the test acts as expected:
```
amy.reset()
amy.load_sample('sounds/partial_sources/CL SHCI A3 LP.wav', preset=1024)
Loaded sample over wire protocol. Preset #1024. 22560 bytes, 11280 frames, midinote 60
amy.send(osc=0, wave=amy.PCM, preset=1024, mode=amy.PCM_LOOP_FOREVER)
amy.send(osc=0, eg0='0,1,700,0')
amy.send(osc=0, note=60, vel=1)   # Starts looping
amy.send(osc=0, vel=0)  # Fades out over 700ms
```
Behavior varies at different pitches, however.  `amy.send(osc=0, note=61, vel=1)` stops after 11 repeats, and `amy.send(osc=0, note=63, vel=1)` stops after one repeat.

I suspect the `terminate_on_silence` flag is engaged, and different pitches are aligning brief silences with the measurement block.  However, `terminate_on_silence` is supposed to be disabled (at least until release) with PCM.

dpwe:
For `PCM` voices, `synth[osc]->phase` holds the sample position within the waveform with S23.8 resolution, i.e. there are 23 integer bits (so 8MSa = 190 sec @ 44 kHz) and 8 fractional bits.

Previously, the per-sample advance was expressed in the same resolution.  When samples were dramatically slowed down, e.g. 3 octaves below "natural", this could lead to 4 or 5 bit per-sample advances, which led to clearly audible quantization of the pitches (since the phase advance determines the overall playback speed of the sample).

This change introduces a new `phase` variable within the `render_pcm` block loop that is shifted left 8 bits, giving a total fractional resolution of 16 bits for per-sample phase advance (and hence frequency).

Only the top 8 bits of cumulated phase is held between blocks (in the S23.8 `synth[osc]->phase` field), but since `AMY_BLOCK` is usually 256 samples, the cumulated phase in `AMY_BLOCK`'s worth of .16-resolution samples is 8.8, i.e. the bottom 8 bits are zero, so nothing is lost by not storing them.  (If `AMY_BLOCK` is, for instance, only 128 samples, we lose one bit - 1/512 of a sample period - between blocks).

Fixes the issue addressed in PR #1050 without limiting the length of PCM samples.

Along the way we also fixed #1007.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1288

PR #1070, opened 2026-08-09
src/pcm.c: *Restarting* a PCM will attempt to delay until a zero crossing to avoid clicks.

dpwe:
#1051 identified a problem: When a particular PCM osc is retriggered (as happens with playing the same GM Drum sound in rapid sequences), the PCM waveform snaps back to the beginning immediately without any kind of transition; this can lead to a hard click if the retrigger happens during a moment of large output amplitude.

These changes address that situation by *delaying* the retrigger by up to 512 samples (i.e., ~11ms at 44 kHz) to allow the waveform to reach a zero crossing; the restart applies directly after the zero crossing, so the hard transition is minimized.

Adding an unexpected, data-dependent delay to commanded note onsets seems a bit questionable, but the delay is small and hopefully not noticeable.  The click, before removal, however, was extremely noticeable.

Thanks to rt-rtos for the original diagnosis.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1293

rt-rtos:
The zero-crossing defer is likely the correct generalized default.
My attempt (#1051) imposed a synthetic compensation signal, chasing zero onset
latency. Cost here is self-targeting: delay scales with the
tail's time-to-zero, so hats pay nothing and only the material that was
going to click pays at all. On a 50 Hz kick tail the search runs near the
full cap, and the payment is 0-11 ms of phase-dependent, uncompensatable
timing jitter on groove-carrying retrigs; a click (partially masked by the
new attack) exchanged for feel (unmaskable). `PCM_MAX_ZERO_SEARCH_LEN` is
effectively the lowest fully-covered frequency (512 = a 43 Hz half-period
at 44.1 kHz; below that the min-|val| fallback keeps a residue of the
click) - could be worth a note at the definition, since it's the knob for
trading delay against coverage.

Two bugs, both reproduced on host sim:

- The `PCM_LOOP_ONCE_INTERNAL` wrap promotes `state` to `next_state` but
  never restores `preset->loopstart/loopend`, so a retriggered looping-mode
  note sustains on the hijacked [restart, old-playhead] region. Repro: loop
  authored [2000, 3000], retrig ~1280 samples in -> osc sits in `PCM_LOOP`
  with `loopstart=0 loopend=1400`, permanently. The one-shot modes never
  read `loopend`, so the drum tests can't see it.

- The chained-osc trigger_phase guard in `play_delta` tests
  `synth[d->osc]->wave` but acts on `synth[osc]` - a non-PCM osc chained
  under a PCM head never reapplies its `trigger_phase` on note-on. Repro:
  identical `trigger_phase=0.25` SINEs chained under a SINE head vs a PCM
  head; on retrig the first snaps to `F2P(0.25)`, the second keeps its
  running phase. Looks like a typo for `synth[osc]->wave`.

Question: for `AMY_PCM_TYPE_FILE`, `pcm_find_zero_crossing` returns -1 and
the file path skips the wrap - as far as I can tell a retrigger of a
streamed sample now never restarts. Intentional?

Happy to send a PR for the first two.

dpwe:
Thanks!  I fixed all these (I think) in #1082

PR #1073, opened 2026-08-10
midi_mappings: take status/channel as arguments, so synths > 16 work

dpwe:
## The problem

The MIDI mapping dispatch recovered the channel from MIDI byte 0, which caps it at 16. `patches.c` therefore had to pack the synth number into a status byte:

```c
bytes[0] = 0x90 + (0x0F & (e->synth - 1));
```

So a drum kit on synth 20 looked up **channel 4's** mappings, and one on synth 0 looked up **channel 16's**. `85bafab` added `TestSynthDrumsChan20` demonstrating the silence that results.

## The change

`midi_message_handler_to_queue()` and `yield_midi_message_handler_events()` now take the already-decoded status nibble and channel, plus a pointer to the data bytes that follow — callers unpack byte 0 themselves:

```c
void midi_message_handler_to_queue(uint8_t status, uint16_t channel, uint8_t *data, uint16_t len, ...);
void *yield_midi_message_handler_events(uint8_t status, uint16_t channel, uint8_t *data, uint16_t len, ...);
```

`patches.c` passes `e->synth` straight through as the channel. `midi_msg_handler()` keeps its raw-bytes signature — tulipcc's `amy_connector.c` calls it — and does the unpacking for the live-MIDI path; `amy_event_midi_message_received()` already had `status` and `channel` in hand and just forwards them.

The mapping root lists were also fixed `[AMY_NUM_MIDI_CHANNELS + 1]` arrays, so storing a mapping for synth 20 was rejected before it could ever be looked up. They're now allocated at init to `max(16, config.max_synths)`, and the eight copies of the range check collapse into one `mapping_channel_ok()`. `AMY_NUM_MIDI_CHANNELS` stays 16 where it belongs — `amy_midi.c`'s active-channel bitmap for real MIDI input.

Two follow-ons in the same area:

- **Channel 0 is now allowed.** It isn't reachable from a MIDI cable, whose channels number from 1, but it's a perfectly good synth number. Previously a patch on synth 0 silently failed to install its note maps and its notes were routed to synth 16.
- **`midi_clear_channel_mappings()` no longer recurses.** Each root list holds only mappings of its own type on its own channel, so clearing a type is just emptying its list — no self-call, no redundant filtered walk, and `midi_active_channel_set()` runs once instead of three times.

## Test fixture correction

`TestSynthDrumsChan20` could not have matched the reference wav it shipped with, for reasons unrelated to channel 20: it ran with `default_synths = False` and loaded `patch=384`, but its reference is `TestSynthDrums` output, and outside a `GAMMA9001` build `amy_default_synths()` puts **patch 258** on channel 10. Different fixture, different kit. Corrected to `default_synths = True` + `patch=258`; the committed reference needed no regeneration.

## Verification

- `make test` — 127/127 pass (was 126 pass / 1 fail).
- `make ctest` — all green.
- Clean build, no new warnings.
- The same kit on synths **0, 10, 20 and 63** renders byte-identical wavs.

One pre-existing quirk worth knowing, unchanged here: loading a drum kit onto synth 0 *with* `default_synths` on is loud, because `amy_default_synths()` has already put the 1-voice sine bleeper there and the 38-voice kit lands on top of it. Measured against the old binary it was worse before this PR (-2.9 dB vs -5.5 dB now).


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1295

PR #1075, opened 2026-08-10
Give each osc two mod sources (COEF_MOD0 / COEF_MOD1)

dpwe:
## Why

An oscillator could name exactly one modulating oscillator. Route an LFO to pitch and there was nothing left to route to amplitude or filter cutoff.

## What

`mod_source` is now a vector of `NUM_MOD_SOURCES` (2) oscs, and CtrlCoefs gain a second modulator column. The first source feeds `COEF_MOD0`, the second `COEF_MOD1`.

```python
amy.send(osc=1, wave=amy.SINE, freq=3,  amp={'const': 0.5, 'vel': 0, 'eg0': 0})
amy.send(osc=2, wave=amy.SINE, freq=11, amp={'const': 0.5, 'vel': 0, 'eg0': 0})
amy.send(osc=0, wave=amy.SINE, mod_source=[1, 2],
         freq={'const': 220, 'mod0': 0.1},   # 3 Hz vibrato
         amp={'const': 1, 'mod1': 0.5})      # 11 Hz tremolo
```

Back-compat is preserved on every surface: `COEF_MOD` is kept as a name for `COEF_MOD0` so existing C reads unchanged, `mod_source=3` still means slot 0, and Python's `'mod'` is an alias for `'mod0'`.

## This is not a breaking wire change, and that was the design constraint

`COEF_MOD1` is appended at index **9**, not slotted in beside `COEF_MOD0` at 6. A stored patch string is a positional list, so inserting a column mid-way would have made every saved `bend` read back as a modulator depth — in this repo's own patch data that was ~385 of 391 patches, and out in the world it would have hit AMYboard World sketches and every user-saved patch.

Nothing that existed moved. The check that this holds: **`src/patches.h` and `src/patches.generated.js` regenerate byte-identical to `main`** — they don't appear in this diff at all.

Coef order on the wire is now `const, note, vel, eg0, eg1, mod0, bend, ext0, ext1, mod1`.

## Implementation notes for review

- **`MOD_SOURCE` becomes a delta range** (`MOD_SOURCE_START..MOD_SOURCE_END`), exactly like `ALGO_SOURCE`. One delta is emitted per *named* slot, so `mod_source=3` leaves slot 1 alone — the same "only what you specify changes" rule the coefs already follow. Wire: `L3` or `L3,4`; Python: `3`, `[3, 4]`, or `[None, 4]`.
- **`mod_osc_would_cause_loop()` needed real thought.** With one source per osc the mod graph was a chain and a `while` loop sufficed; with two it is a DAG, so the check now branches. It is bounded by a node-visit budget (`NUM_MOD_SOURCES * AMY_OSCS`) rather than by recursion depth, so a deep diamond can't make cycle-checking blow up exponentially. Exhausting the budget rejects the assignment, which is the safe direction.
- Param ids after `CHAINED_OSC` shift, as they already do whenever `NUM_COMBO_COEFS` changes (`DUTY = AMP + NUM_COMBO_COEFS`, etc.). They're internal to deltas and never persisted.

## Positional vectors cleaned up

Since positional coef vectors are what made this risky, the ones in our own source are now dicts: `juno.py`'s `_freq_coef_string` → `_freq_coef_coefs`, one test, the `parse_ctrl_coefs` docstring, and the docs examples. These emit identical wire strings — that's part of why `patches.h` matches.

Two bindings had no safe form to switch *to*, so they got one:
- **Godot** `_format_ctrl` accepted only numbers/strings/arrays — no dict at all. It now takes a Dictionary, and `COEF_FIELDS`/`COEF_ALIASES` are **generated** from the Python rather than hand-copied.
- **JS** got the same `mod` → `mod0` alias Python has, so existing `{mod: …}` keeps working.

## Tests

`make test` 129/129, `make ctest` all green, clean build, all generator `--check`s pass.

Two new tests, whose references I verified measure what they claim rather than just accepting current output:
- `TestTwoModSources` — 3 Hz → `mod0` → vibrato and 11 Hz → `mod1` → tremolo on one osc; both rates confirmed present in the reference.
- `TestSecondModSourceOnly` — `mod_source=[None, 2]` must leave slot 0 unset rather than shifting osc 2 into it; vibrato swing collapses while the `mod1` tremolo persists.

Both were generated under an earlier index layout and still match byte-for-byte after moving `COEF_MOD1` from 6 to 9, because they're written in dict form. That's the dict-vs-positional argument demonstrated rather than asserted.

## Downstream (tulipcc), not in this PR

Checked and staged locally, to land with the pin bump:
- `tulip/shared/amy_connector.c` carries a **hand-written** copy of the coef names that `tulip.amy_message()` walks — updated, plus an alias table. Impact was mild (unknown key → fall back to the Python path, correct but slower), never wrong output.
- `AMY_AGENTS.md` guidance updated for `mod0`/`mod1`.
- `spss.js` did `Number(event.mod_source)`, which is `NaN` for `"1,2"`.
- Pin-gated and CI-enforced by `amy-pin-check.yml`: `tulip/shared/amy_kwmap.h` needs `mod_source` `'I'` → `'L'`, and `tulip/server/refdocs/amy/` needs the `synth.md`/`api.md` edits. Both scripts source from the pinned commit, so they can only be refreshed once this merges.

Separately, tulipcc `main` does not currently build against amy HEAD — `user_c_dsp.c` uses the removed `AMY_NUM_BUSES` and the bus-postprocess hook is still `uint8_t bus`. Pre-existing pin-gap fallout from the max-buses work, unrelated to this PR, but it will need fixing as part of whichever pin bump carries this.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1302

PR #1079, opened 2026-08-10
tulipcc-pin: regenerate amy_kwmap.h too, not just refdocs

dpwe:
The auto-pin workflow refreshes `tulip/server/refdocs/amy` for the new pin but not `tulip/shared/amy_kwmap.h`, so every pin PR it opens fails tulipcc's `amy-pin-check`:

```
STALE: tulip/shared/amy_kwmap.h does not match _KW_MAP_LIST at amy 834c7ea845eb
Run: python3 tulip/shared/gen_amy_kwmap.py  (then commit)
```

## This is not occasional

The generated header stamps its source SHA in its first comment line:

```c
/* GENERATED by tulip/shared/gen_amy_kwmap.py -- do not edit.
 * Source of truth: amy/amy/__init__.py _KW_MAP_LIST at 834c7ea845eb
```

so it is stale after **every** pin bump — even one that changes nothing in `_KW_MAP_LIST`. `amy-pin-check` has enforced it since 2026-07-28, the commit that introduced the header, and not one of the four bot bumps since (`amy#1067`, `amy#1045`, `amy#896`, `amy#1012`) has touched it. Each needed a human to run the generator before the PR could go green.

tulipcc#1302 is the case that made it obvious. amy#1075 changed `mod_source` from an int to a list (`'I'` → `'L'`), so besides being red, a stale kwmap would have sent every message carrying a `mod_source` down the Python fallback instead of the C fast path.

## The fix

`gen_amy_kwmap.py` already takes `--sha`, exactly like `sync_amy_docs.py`, so it runs in the same clone against the same merged SHA — no new plumbing.

- Both scripts run in one `if`, kept **all-or-nothing** with the docs sync. A half-refresh fails CI just the same, so there's nothing to gain by keeping one script's output when the other fails.
- The tree-building step now filters against an allow-list of the two derived paths rather than one hardcoded prefix, so a script that ever starts touching something else can't smuggle an unrelated file into the bump commit.
- `refdocs_changes.txt` → `derived_changes.txt`, and the commit-message note names both files.

No change to the `pull_request_target` safety posture: it still only clones `shorepine/tulipcc@main` (trusted) and reads amy via GitHub raw at the merged SHA. It never checks out or executes the PR's code.

## Verified

Ran the new step **verbatim** against a fresh clone of `tulipcc@main` (which pins `834c7ea`) with `AMY_SHA=50ec2e3`, amy main's current tip — i.e. the real next-pin scenario. It emits:

```
M	tulip/server/refdocs/amy/_VENDORED_FROM.txt
M	tulip/server/refdocs/amy/synth.md
M	tulip/shared/amy_kwmap.h
```

The old step emitted only the first two. The next auto-pin PR should now open green.

### DIFF
--- a/.github/workflows/tulipcc-pin.yml
+++ b/.github/workflows/tulipcc-pin.yml
@@ -154,33 +154,44 @@ jobs:
             core.setOutput('merger', merger);
             core.setOutput('amy_pr', String(pr.number));
 
-      - name: Refresh tulipcc's refdocs/amy snapshot for the new pin
+      - name: Refresh tulipcc's amy-derived files for the new pin
         env:
           AMY_SHA: ${{ steps.gather.outputs.amy_sha }}
         run: |
-          # tulipcc CI's refdocs-fresh check fails any PR that moves the amy gitlink
-          # without regenerating tulip/server/refdocs/amy (its _VENDORED_FROM.txt
-          # stamps the source SHA). Run tulipcc's own sync script against the new pin
-          # and hand the changed files to the next step to fold into the bump commit.
+          # tulipcc's amy-pin-check fails any PR that moves the amy gitlink without
+          # regenerating BOTH files it derives from the pinned commit:
+          #
+          #   tulip/server/refdocs/amy/   (sync_amy_docs.py; _VENDORED_FROM.txt
+          #                                stamps the source SHA)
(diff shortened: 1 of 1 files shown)


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1304

PR #1081, opened 2026-08-11
Make reset=RESET_TIMEBASE render-thread-safe

bwhitman:
## The bug

`amy_reset_sysclock()` ran in the wire parse, on whatever thread called `amy_add_message()`. It zeroed `total_blocks` / `total_samples` / `sequencer_tick_count` and re-anchored `next_amy_tick_us` while the render thread could be inside `sequencer_check_and_fill()` — which had already computed `now_us` against the **old** clock and does an unguarded read-modify-write of `next_amy_tick_us` (the hazard sequencer.c's own thread-safety comment names). The catch-up `while (now_us >= next_amy_tick_us)` loop then saw a stale `now_us` seconds ahead of the freshly re-anchored deadline and replayed thousands of ticks in one block.

Measured on the Tulip desktop build: with live periodic sequencer events, a `reset=RESET_TIMEBASE` sent at ~4 s of uptime smears the pattern into a continuous roll for a second or more. Severity varies run to run — it is a race.

## The fix

- `amy_reset_sysclock()` only sets a pending flag, under the lock.
- The render thread applies the reset **between blocks**, at the top of `amy_fill_buffer()` — the one per-block function only the render path calls (`amy_execute_deltas()` is also reachable from parse threads via patch/transfer loads, so applying there would reintroduce the race). Applied on the same thread that does both racy read-modify-writes, the zeroing can no longer interleave with them.
- Applying it also **rebases queued deltas onto the new timeline**, so their relative timing survives: a delta due 200 ms out is still due 200 ms out, one already due plays on that block. (`RESET_EVENTS` remains the way to drop them.)

A sequenced `S16384` (a reset fired from a stored wire, i.e. from inside the tick loop itself) is also safe: the request no longer touches `next_amy_tick_us`, so the in-flight catch-up loop finishes on the old timeline and the zeroing lands at the block boundary.

## What this revision drops

An earlier version of this PR also made a *requested but not yet applied* reset read as though it had already happened: `amy_sysclock64()` split into a public and an internal `_raw()` variant, `sequencer_ticks()` returning 0 while the flag was up, and `sequencer_add_wire()` judging "already due" against the not-yet-running timeline. That machinery is gone — it wasn't worth its weight now that the API rarely asks callers to read the clock at all.

**Correction to an earlier revision of this description:** it said the reset now honours `time=`. `amy.send()` has no `time` kwarg (no entry in `_KW_MAP_LIST`, no `t` wire code) — the schedulable field is `ticks=`, plus `amy_event.time` for C callers. The docs in this PR say that now.

Moving the delta rebase from *request* time to *apply* time is what makes dropping it safe for events: the rebase now runs against the clock as it reads when the reset actually lands, so deltas queued in the gap between the message and the next block boundary are rebased too. Those were exactly the events the masked getters existed to protect.

**Given up deliberately:** for the few ms until the next block, `amy_sysclock()` and `sequencer_ticks()` still report the old timeline, and a sequencer one-off scheduled in that window is judged due against the old tick count. Both are documented in `api.md` / `synth.md` rather than papered over.

Net effect on the diff: `src/api.c` and `src/sequencer.c` are no longer touched at all.

## Also in this PR

- **`RESET_TIMEBASE` is an ordinary event.** It was handled only in the wire parse, so `amy_add_event()` with `reset_osc = RESET_TIMEBASE` was silently dropped (`play_delta` had no branch for the bit) and `amy_reset_sysclock()` was the only way in from C. Now that applying the reset is deferred, "handling" it is just raising a flag — safe from any thread — so it rides a delta like every other reset bit and can be scheduled with `ticks=`. `RESET_AMY` and `RESET_EVENTS` stay in the parse: restarting AMY, and emptying the queue a delta would sit in, can't be carried in an event.
- **Bug found on the way:** that parse branch unset the *whole* `reset_osc` field once it had handled its own bits, so anything combined with them never reached `play_delta` — `RESET_EVENTS|RESET_ALL_OSCS` emptied the queue and left every oscillator untouched. It now clears only the bits it handled.
- **`RESET_SYNTHS` is a deprecated alias for `RESET_ALL_OSCS`.** Both always called the identical `amy_reset_oscs()`; the only difference was parse-time vs delta. The constant stays so `S262144Z` in saved sketches and the AMYboard web editor keeps working. One behavior is given up: patch loads are *not* scheduled (they run synchronously in `amy_event_to_deltas_queue`, ignoring `latency_ms` and `amy_event.time`), so at `latency_ms > 0` a teardown sent just before a patch load now lands after it. At the default zero latency the order still holds, because `flush_due_deltas()` runs immediately before every load. Measured with the AMYboard boot pattern: 6 voices at latency 0, 0 voices at latency 100. Nothing in amy or tulipcc sets `latency_ms`.

## Testing

- New `tests/test_timebase_reset.c` (wired into `make ctest`) pins the contract: the reset lands at a block boundary (clock still old right after the message, restarted one block later); a queued event keeps its distance from now (~208 ms measured for a 200 ms schedule); an already-due event isn't lost. Verified the middle one **fails without the rebase** — the event never plays. The third passes either way, since `flush_due_deltas()` plays it on the parse side before the reset lands; its comment says so rather than implying more coverage than it has.
- Three more cases for the event-path change: the reset works through `amy_add_event()`, a reset scheduled with `amy_event.time` waits for its time, and combined reset bits all take effect. All three fail without it. Plus one pinning `RESET_SYNTHS` as a working alias, including that a patch load sent right after a teardown still survives it.
- `make ctest` 8/8, `make test` 132/132, on a forced-clean rebuild.


dpwe:
I feel like just letting `RESET_TIMEBASE` execute as part of `play_delta` would mostly fix this.

 Re-basing the pending scheduled events is interesting. I'm not sure exactly what behavior we'd want in a situation where some future events have been scheduled at specific times, but then the timebase is reset.  Either they should all be abandoned (but that's what `RESET_EVENTS` is for), or they could be retained with their original absolute due-times (now re-interpreted under the new timebase), or they can preserve their relative times (as this code implements).  The latter seems like the least likely to confuse users.

Changing `RESET_TIMEBASE` to parse-time was applied in 7e55ec693c235afde76d27a955f0db2ba26c7208 (in 2024); unfortunately, I don't know what problem that was solving, so I'm not sure what to look for if we change it back.  It's possible that the timebase-rebasing added in `reset_sysclock` here is enough to solve the problems that the original "reset timebase at parse" was attempting to solve, and would be a more architecturally-consistent fix.

dpwe:
I discussed this with Claude who convinced me this is the best approach.  The original move of RESET_TIMEBASE to the event-writing thread was to deal with situations like `amy.send(reset=amy.RESET_TIMEBASE); now = amy.millis()` where the read of the AMY `sysclock` might happen before the `RESET_TIMEBASE` delta had been executed. I think the core problem bothering me is the thread sending the commands wants to know the "current time" so it can schedule, but "current time" doesn't really exist, for instance in situations where everything is being stacked up in the delta queue.

Maybe a deeper fix would be to provide scheduling which is "now-relative".  I think I proposed this once (using negative times to indicate relative) but it was seen as too confusing.  I think in reality, users who are scheduling absolute-time events always want them anchored on "now" (i.e., no-one really wants to schedule something at a precise time after the most recent timebase reset, unless they performed that timebase reset themselves), but I think the problem is that if all scheduling is relative to "now", and "now" is moving in real-time, a sequence of scheduled commands that are designed to have a fixed timing relative to each other can end up drifting if "now" changes while they are being scheduled.

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1323

PR #1085, opened 2026-08-12
Allow fractional midinote on memory PCM presets

bwhitman:
A sampled instrument's per-sample tuning correction is fractional — an Akai S1000 sample rooted 4 cents sharp of C4 is midinote 60.04 — but the memory-PCM preset midinote was `uint8_t` end to end, so the fraction was silently truncated and callers had to re-apply the correction on every note-on instead of baking it into the preset. (Found while teaching CPython AMY to play Akai sample CDs; multisample keymaps carry these per-sample tune offsets everywhere.)

The pitch math was already float-ready (`logfreq_for_midi_note()` takes a float, and `preset->midinote` was just promoted at the call site in `render_pcm`), so this only widens storage and transport:

- `memorypcm_preset_t.midinote` and `pcm_load()` take a `float`
- the `load_sample` (`z`) and `start_sample` (`zS`) wire messages parse the midinote field as float via a new `parse_field_float()` helper; the other five fields keep their integer parse, and integer midinotes parse exactly as before (the uint32 list parser already skipped a fractional token to the next comma, so mixed senders/receivers stay compatible in both directions)
- the Python senders format midinote with `%g` instead of `%d`

The baked-ROM/gamma tables (`pcm_map_t`) keep `uint8_t` midinote — that's a generated data format and it widens on assignment.

**Verification**
- Bit-exact equivalence: `load_sample(..., midinote=60.5)` + `note=60` renders identically to `load_sample(..., midinote=60)` + `note=59.5`, and differs from the truncated behavior
- New test `TestPcmMemoryFractionalMidinote` with reference WAV
- All 124 existing tests pass unchanged; `make check-c-api` clean


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1307

PR #1086, opened 2026-08-12
Allow stereo memory PCM presets over the wire

bwhitman:
Stacked on #1085 (branched from it; GitHub will retarget to `main` when that merges).

The engine has supported stereo PCM presets all along — `render_pcm` handles interleaved frames and `wave=PCM_LEFT`/`PCM_RIGHT` selects a channel (`PCM` mixes both) — but only file-streamed presets (`disk_sample`) and bus sampling (`zS`) could actually be stereo. The `load_sample` (`z`) wire path hardcoded `channels=1` and the Python senders downmixed stereo input before transfer. (Found while teaching CPython AMY to play Akai sample CDs, where everything ships as stereo multisamples.)

- the `z` message takes an optional 7th field, `channels`: absent/0/1 is mono (bit-identical to the old message layout), 2 means the transfer payload is interleaved L/R frames
- `amy.load_sample_bytes(stereo=True)` now loads the interleaved input as a stereo preset instead of silently keeping only the left channel (no in-tree caller used `stereo=True`)
- `amy.load_sample()` grows `stereo=True` to keep both channels of a stereo WAV; the default stays the historical first-channel downmix, and transfer chunks stay 188 bytes (94 mono / 47 stereo frames)
- `docs/api.md`'s `z` row documented a `channels` field that never existed, in a position that didn't match the code — it now describes the actual layout

**Verification**
- New test `TestPcmMemoryStereo` plays `sounds/220_440_stereo.wav` (440 Hz left / 220 Hz right) through `PCM_LEFT`/`PCM_RIGHT` on hard-panned oscs, then through `PCM` as a mix; FFT of the render confirms 440 stays left, 220 stays right, and the mix carries both at equal energy
- All existing tests pass unchanged (132 total)


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1308

PR #1087, opened 2026-08-12
Audit docs/api.md against the wire parser

bwhitman:
Stacked on #1086 (which is stacked on #1085); GitHub will retarget as those merge.

Checked every row of the wire-command tables in `docs/api.md` against `parse.c` (treated as the source of truth) and the `amy_event` struct. **No wire commands were missing from the doc**, but several rows were wrong, and every C-column **TODO** is now resolved.

**Corrections**
- `P` sets `amy_event.trigger_phase` (not `.phase`); `m` sets `.portamento_ms` (not `.portamento`)
- `g` (Alles client) and `W` (Tulip CV) are not parsed by AMY core at all — their rows are removed entirely (the table gave them `amy_event` fields that don't exist)
- `H` (ticks) works only as a wire-message prefix; `amy_event.ticks[3]` exists but `amy_add_event` never reads it — C hosts schedule with `e.time`, now documented with an example
- `zI`/`zB` are recognized only in the MIDI **sysex** layer (`amy_midi.c`), never by the wire parser; `zD`/`zP`/`zI`/`zB` had Python kwarg names (`dump_sysex`, `exec`, `ping`, `reboot`) that don't exist in `amy.py`
- wave list was missing `SILENT` and capped at 21 (`WAVE_OFF` is 22); `reset` now lists all handled flag bits and which of them act at parse time; `synth` range is 0..`max_synths`-1 (default 64), not 0-31; memory-patch IDs follow `max_memory_patches`

**The TODOs**
Those cells marked commands that will never be `amy_event` fields — they're *immediate commands*, not schedulable parameters. Rather than invent struct fields, the C column now names the exported C function each command calls (`pcm_load`/`pcm_unload_preset`, `pcm_load_file`, `start_receiving_sample`/`stop_receiving_sample`, `sequencer_midi_start`/`_stop`, `amy_external_midi_sync`, `show_debug`, `patches_store_patch`, `midi_store_mapping`), so a host linking AMY can call them directly. A new intro paragraph explains the two kinds of commands, and the C API section gains examples for `e.time` scheduling and direct `pcm_load()` sample loading (no base64 transfer needed in C).

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -22,6 +22,36 @@ amy_add_event(amy_event *e);
 amy_add_message(char *message);
 ```
 
+To schedule an event for the future, set the `time` field (milliseconds on
+AMY's clock; see `amy_sysclock()`). This field only exists on the C struct --
+wire messages play as they arrive, and the sequencer (`ticks`) handles
+musical-time scheduling:
+
+```c
+amy_event e = amy_default_event();
+e.osc = 0;
+e.midi_note = 60;
+e.velocity = 1;
+e.time = amy_sysclock() + 1000;   // play one second from now
+amy_add_event(&e);
+```
+
+C hosts that link AMY directly can load PCM presets without the wire
+protocol's base64 transfer: `pcm_load()` allocates the preset and returns its
+sample buffer for you to fill (this is what the `z` wire message does
+internally):
+
+```c
+// frames*channels int16 samples; channels=2 means interleaved stereo,
+// played with wave=PCM_LEFT / PCM_RIGHT (or PCM to mix both channels)
+int16_t *buf = pcm_load(preset, frames, samplerate, channels,
+                        midinote /* float, may be fractional */,
+                        loopstart, loopend);
+memcpy(buf, my_frames, frames * channels * sizeof(int16_t));
+// ... play with wave=PCM, preset=preset.  Remove with:
+pcm_unload_preset(preset);
+```
+
 Two sample types appear in this API: `output_sample_type` is a final
 interleaved audio sample, an `int16_t`; `SAMPLE` is AMY's internal sample
 format, S8.23 fixed point in an `int32_t` (so 1.0 is `1<<23` — see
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -400,18 +439,16 @@ Default AMY has 4 buses, 0..3.  Set `max_buses` in `amy_config_t` before `amy_st
 
 | Wire code   | C `amy_event` | Python / JS   | Type-range  | Notes                                 |
 | ------ | -------- | ---------- | ----------  | ------------------------------------- |
-| `H`    | `ticks[3]` | `ticks` | int[,int[,tag]] | Tick, period, tag for sequencing (see "AMY's sequencer" in synth.md). `tag` omitted: stored but not individually cancelable. `period` also omitted: a one-off event at that tick. **If used in a wire string message**, the `H` **must** be the first character of the message. |
+| `H`    | wire only (schedule C events with `e.time`) | `ticks` | int[,int[,tag]] | Tick, period, tag for sequencing (see "AMY's sequencer" in synth.md). `tag` omitted: stored but not individually cancelable. `period` also omitted: a one-off event at that tick. The `H` **must** be the first character of the message; the rest of the message is stored as the sequenced payload. (The `amy_event.ticks[3]` field exists but is not read by `amy_add_event` -- from C, send a wire string.) |
```
dpwe: This is strange.  The PR comment says `amy_add_event` doesn't look at e.ticks[] but it's literally the first thing it does (at `api.c:309`) - if ticks are present, it goes to great lengths to rewrite the event as a wire command starting with `H` so it is correctly handled.  Please DON'T encourage C users to use the e.time field, this is reserved for internal near-term sequencing and ideally would not be exposed to C users at all.

docs/api.md
```diff
@@ -22,6 +22,36 @@ amy_add_event(amy_event *e);
 amy_add_message(char *message);
 ```
 
+To schedule an event for the future, set the `time` field (milliseconds on
```
dpwe: Please no.  You can show how to use `e.ticks[3]` to mirror `H,,` in the wire code instead.

dpwe (changes requested):
`e.ticks[3]` works exactly like `ticks='a,b,c'` in Python, please do not suggest that the user use the `e.time` field, I am trying to get rid of it.

docs/api.md
bwhitman: You're right, I misread this — `amy_add_event` checks `e.ticks[]` first thing and serializes to an `H` wire message. Fixed in e08387a: the `H` row's C column is back to `ticks[3]` with a note that it works exactly like Python's `ticks=`, and `e.time` is no longer mentioned anywhere in the doc.

docs/api.md
bwhitman: Done in e08387a — the intro example now schedules with `e.ticks[TICKS_TICK] = sequencer_ticks() + AMY_SEQUENCER_PPQ` (mirroring wire `H`/Python `ticks=`), noting `TICKS_PERIOD`/`TICKS_TAG` for the repeat/cancel forms. All `e.time` references removed.

PR #1088, opened 2026-08-12
A patch load keeps the synth's bus, and the patch's FX land on it

bwhitman:
Two fixes for how patch loading interacts with buses, found from Tulip — a drum machine on its own bus picked up its neighbor's Juno chorus after a sample change.

**1. A re-patch keeps the instrument's bus.** Loading a patch rebuilds the voices' oscs (reset to bus 0) and re-registers the instrument, so a patch change that didn't re-state the bus dropped the synth onto the default bus, into everybody else's FX. Measured from Tulip through a zeroed bus volume: a fresh synth reads 0.000 (on its bus), after a bare re-patch 0.052 — audible, on bus 0. `patches_load_patch` now inherits the standing bus when the event doesn't name one, and the existing pipeline (`instrument_add_new`, the per-osc BUS deltas fanned out by `patches_event_has_voices`) re-routes exactly as an explicit bus would.

**2. A patch's own FX settings land on the loading synth's bus.** 127 of the ROM Juno patches carry chorus config (`k1`/`k0`) in `v` messages with no synth attached, so those bus-directed params fell to bus 0: chorus configured on a bus the loading synth isn't on colors whoever IS there, and leaves the patch itself dry. Live loads (the global delta queue) now aim them at the synth's bus; a patch being STORED keeps its params unresolved, to be aimed at whichever synth later loads it.

**Tests**: both covered in `tests/test_bus_config.c` (`make ctest`) — all four new checks fail without the fixes. `TestSynthBusCmds`' ref is regenerated: it loads a chorus-carrying Juno patch on synths on bus 0 and bus 1, and each patch's chorus now lands on its own synth's bus, which is the intended change in rendering. The other 129 audio tests pass byte-identical.


src/patches.c
```diff
@@ -843,6 +843,21 @@ void parse_patch_string_to_queue(char *message, int base_osc, struct delta **que
         }
         pos = yield_event_from_message(message, &e, pos);
         if (pos > 0) {
+            // A patch's own FX settings (a Juno patch's chorus, "k1" --
+            // 127 of the ROM Juno patches carry one) ride in 'v'
```
dpwe: I don't know what this means.  The Juno patches in `patches.h` all have the same pattern: A bunch of `Z`-separated, `v`-directed commands setting up the oscs, and a final global FX command phrase starting with `x` and including `k` (NOT part of a `v`-rooted phrase). 

If you set the synth, bus-directed commands will, I believe, pick up the bus from the synth.  So please just do "if (event_addresses_bus(&e)) e.synth = synth;"

dpwe (dismissed):
I would prefer to ensure that the FX commands in the patch string end up on the right bus by setting their synth field.

src/patches.c
bwhitman: Done in 01a66bb — replaced the hand-rolled bus resolution with `if (event_addresses_bus(&e)) e.synth = synth;` as suggested. The bus then resolves from the synth in `amy_event_to_deltas_queue`, and `patches_event_has_voices` early-exits for the pure-FX phrase so nothing else changes. Both ctest checks still pass and `make qtest` is bit-exact against the ref (130 pass).

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1309

PR #1089, opened 2026-08-12
Harden the wire-message parser against arbitrary junk input

bwhitman:
## Summary

Base64 payload chunks from a desynchronized sample transfer leaking into `amy_add_message()` could segfault the engine (observed as SIGSEGV / exit 139 in the CPython `c_amy` module). Arbitrary text reaching the parser should never crash AMY — at worst it should complain and drop the message. Fuzzing the wire parser with random base64- and command-shaped strings reproduced the reported crash and surfaced ten more crashes/hangs, all fixed here.

## The reported segfault

`patches_load_patch()`'s user-patch branch indexed `memory_patch_oscs[]` / `memory_patch_deltas[]` with no upper bounds check. Junk messages storing patches until the pool was full made `patches_store_patch()` auto-assign an out-of-range number (the "patch number 1056 is out of range (1024 .. 1056)" printed just before the crash), leave it in the event, and the subsequent load read a garbage delta pointer → SIGSEGV. Also reachable with an explicit out-of-range `K` number, or a patchless instrument whose patch number reads back as the UNSET sentinel.

## Everything fixed

**Segfaults**
- `patches_load_patch`: bounds-check user patch numbers before indexing (patches.c)
- `patches_store_patch`: don't leave a rejected auto-assigned patch number in the event (patches.c)
- `ensure_osc_allocd`: bounds-check the osc number centrally — wire `algo_source` values and base_osc arithmetic could index `synth[]` far out of range; callers already treat false as drop-the-delta (amy.c)
- `set_event_for_osc`: state dumps (wire `D`) walked voice oscs that were never lazily allocated → NULL deref (patches.c)
- `config_reverb` / `config_chorus`: a negative wire level stored as nonzero without allocating delay lines; a later positive level (or the render loop's `bzero(delay_mod)`) then dereferenced NULL. Gate on the allocation itself and clamp negative levels (amy.c)
- `patches_voices_for_load_synth`: `num_voices` > MAX_VOICES_PER_INSTRUMENT overflowed the `voices[]` stack array, then `abort()`ed in `instrument_init` (patches.c)

**Hangs (infinite loops)**
- `midi_mapping_from_message`: an empty payload (`ic`/`io` at end of message) returned -1, exactly cancelling the outer parser's advance (parse.c)
- `log2_lut`: zero/negative (and float-build inf) inputs never exited the normalization loops — negative freq or velocity from the wire hung the render thread (log2_exp2.c, plus `logfreq_of_freq` guard in amy.c)
- garbage tempo (`j0`, `j-1`, huge) made `us_per_tick` 0 and the sequencer catch-up loop spun forever; reject bad tempo at the delta and floor `us_per_tick` (amy.c, sequencer.c)
- `midi_out` on macOS called into CoreMIDI with no client/port when wire `zC2` enabled clock-out without MIDI configured, blocking every tick (macos_midi.m)

**Aborts**
- `wave=CUSTOM` with no custom oscillator registered hit `assert(custom_osc != NULL)`; render silence instead (custom.c)

## Test

`TestFuzzWireParser` in `amy/test.py`: 19 directed repros (one per bug above) plus a deterministic-seed 2000-message fuzz loop (pure base64 noise, command-shaped noise, corrupted valid messages, extreme numeric arguments), rendering a block per message. Afterwards the engine is restarted and must still render an audible sine. Runs in a temp dir (junk `zT` messages legitimately open transfer files) with stderr silenced for the duration.

- `make qtest`: 131 tests pass (bit-exact refs unchanged)
- `make ctest`: passes
- ~100k additional fuzz messages across 26 corpora pass in standalone C harnesses, both with and without rendering; the original `amy.send_wire()` fuzz loop passes on CPython


src/log2_exp2.c
```diff
@@ -12,7 +12,13 @@ static inline SAMPLE lut_val(SAMPLE frac, const LUTSAMPLE *table, const int log2
 
 
 AMY_IRAM_ATTR SAMPLE log2_lut(SAMPLE x) {
-    // assert(x > 0);
+    // Wire-derived params (freq, velocity, envelope levels) can be zero or
+    // negative (or NaN/inf in the float build); the normalization loops below
+    // never terminate for those.  Clamp instead of hanging.
+    if (!(x > 0)) return F2S(-24.0f);  // ~log2 of the smallest positive SAMPLE.
```
dpwe: I'm not crazy about this in principle.  I'm OK with extra checks on arguments during config time to harden behavior, but this is meant to be a very lean calculation called multiple times during rendering, so adding extra checks is a serious matter.  I haven't encountered this routine hanging in regular use; if it only happens when there is random junk fed to the message parser, I'm OK with the system hanging rather than slowing down the rendering by 0.1%

dpwe (dismissed):
I would rather not add any overhead to log2_lut.  I understand it then poses the danger of hanging under some fuzzed configurations, but I don't see that risk as worth slowing down the common case.

src/log2_exp2.c
bwhitman: Fair enough — removed in 4ca5c93: `log2_lut` is back to its original form, zero added instructions. It turns out the boundary guard alone covers everything the fuzzer found: `logfreq_of_freq` now tests `freq<=0` where it previously tested `freq==0` (same single compare, no new cost), and every other `log2_lut` caller already clamps its input (`map_60dB_to_01f` is called with `MAX(0, val)` and returns early at 0; the envelope DX7/exponential paths clamp with `MAX(EPS)`/`MIN(1.0)`). Re-ran the full fuzz corpus (~100k messages, 24 corpora) plus the directed repros against the reverted `log2_lut` — no hangs, and `make qtest`/`ctest` pass.

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1319

PR #1091, opened 2026-08-12
SYNTH_FLAGS_NO_NOTE_WARNINGS: per-synth opt-out of note on/off warnings

bwhitman:
Reworked from the earlier AMY_PRINTF version after feedback — this is now the minimal shape: **one new `synth_flags` bit, riding the existing plumbing** (wire `if`, `instrument->flags`). Nothing changes for any synth that doesn't set it.

`SYNTH_FLAGS_NO_NOTE_WARNINGS` (8): a synth with it set skips the three note-bookkeeping warnings — `note off ... does not match note on`, the forgotten-pool overflow, and note-on-for-note-0. All other diagnostics everywhere are untouched.

**Why:** a sequencer joining a running bar mid-cycle plays note-offs for note-ons that never fired — by design, once per join and once per live edit. On a host whose stderr is a 115200-baud UART those prints land in the audio path at ~4 ms a line against the 5.8 ms block budget: an audible dropout per diagnostic, measured on Tulip's ESP32-P4 bench. Tulip sets the flag on the synths its sequencer apps allocate; hand-played synths keep warning.

Verified: clean `make amy-example`, zero warnings.


bwhitman:
[brian here - i know we already made a thing like this and closed it already after we measured. BUT: that was on another board with a 4Mbps UART connection, and it wasn't taking any time. Actual in service 115200 uart is taking 4ms to printf "note off has no matching note on." and it actually is slowing down AMY in real life. "But brian, why are you triggering that message to begin with?" It's a music player app with APIs called by humans who do things like change the transport head while things are playing.

bwhitman:
ALSO, and i know it's not happening today, but one day we could route printf style messages over another bus, like SYSEX or something, and this is a step towards doing that.

dpwe:
I still don't appreciate "just silence all the warnings" -- we put them there to warn people, not because "oh it doesn't matter".  

If there's a use-case where we expect to send note-offs without corresponding note-ons (as you describe), I would prefer a narrower solution (like a per-synth flag to suppress warning specifically and only for excess note-offs).  

Dropped events and out-of-range args don't seem like great things to just leave uncommented-upon.

bwhitman:
OK, done. The SYNTH_FLAGS_NO_NOTE_WARNINGS fix it for my current problem.

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1312

PR #1093, opened 2026-08-12
Remove leftover osc==999 debug print from hold_and_modify

bwhitman:
`hold_and_modify` carries a hard-coded `if (osc == 999)` that prints a 280-char diagnostic to stderr **every block** for any audible osc numbered 999.

Found the hard way on Tulip's ESP32-P4: its allocator hands raw oscs out top-down from the 1024 pool, so the third app to take raw oscs lands on 999, plays a note, and the print alone (280 chars at 115200-baud stderr ≈ 24 ms against the 5.8 ms block budget) trips the overload failsafe. Measured: three bare sine oscs at 998–1000, render_load 1.4–1.9, `AMY: CPU overload (render 11230 us > 5687 threshold), resetting synth` — entirely printf.

Deletion only; clean `make amy-example`, zero warnings.

### DIFF
--- a/src/amy.c
+++ b/src/amy.c
@@ -1865,15 +1865,6 @@ void hold_and_modify(uint16_t osc) {
     msynth[osc]->feedback = synth[osc]->feedback;
     msynth[osc]->resonance = synth[osc]->resonance;
 
-    if (osc == 999) {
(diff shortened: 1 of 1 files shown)


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1311

PR #1096, opened 2026-08-13
patches: memory-patch slots are recycled, and a dry pool cannot crash

bwhitman:
Four faults in the `patch_string` slot lifecycle, found from tulip where every gamma9001 / exported-beat channel is a `patch_string` instrument. Reported from the bench as a micropython segfault after `patch number 1056 is out of range (1024 .. 1056)`, plus `instrument_number 255 is out of range 0..64 (get_num_voices)` spam on every app quit — with no patch changing involved at all.

## The leak

Every auto-assigned store burned a fresh slot (`next_user_patch_index` only ever grew), so `max_memory_patches` sends — 32 reconfigures of one synth, or a few configure/release cycles of an 8-channel app — emptied the pool for the rest of the session.

Now:
- a store **reuses the addressed instrument's own auto-assigned slot** (a reconfigure — a sample change, a re-sent config — stays on one slot),
- then any **free slot**,
- and **releasing an instrument** (`num_voices=0`) frees its auto-assigned slot.

An **explicitly numbered** patch is never recycled behind the caller — it may be loaded again by anyone who knows the number. Tracked by a per-slot auto flag carried in the `patches_init` block.

## The crash

Once the pool was dry, the refusal was not the end of it: `patches_store_patch` dropped the store, but the event still carried the refused number into `patches_load_patch`, which indexed the 32-slot tables with it **unchecked** and walked a garbage delta pointer — a segfault on the desktop, a Guru Meditation on a device. The user-patch branch range-checks before indexing now.

## Replace, not append

A re-store to the same number appended to the standing delta list (`parse_patch_string_to_queue` appends), silently merging two patches into one slot. A re-store resets the slot first.

## The 255 print

Releasing an instrument printed `instrument_number 255 is out of range 0..64 (get_num_voices)`: the release path unsets `e->synth` on purpose so the rest of the event is not forwarded, and `patches_voices_for_event` then read the unset sentinel into `instrument_get_num_voices`. On a device stderr is a blocking UART line inside the audio path, once per released synth. Guarded.

## Tests

`tests/test_patch_slots.c`, wired into `make ctest`, pins all four: 40 reconfigures stay on one slot, 40 configure/release cycles leak nothing, an explicit slot survives its instrument's release, a re-store replaces, a release prints nothing, and the dry-pool store+load survives.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,7 +105,7 @@ amy-message: $(OBJECTS) src/amy-message.o
 # Plain C tests for things the audio-rendering suite can't reach -- e.g. clock
 # rollovers 50 days out, which you can only hit by fast-forwarding the counters.
 CTESTS = tests/test_clock_wrap tests/test_sequencer_active tests/test_sequencer_bounds \
-         tests/test_bus_config
+         tests/test_bus_config tests/test_patch_slots
 
 # Static pattern rules, so these win over the generic %.o: %.c above (which
 # would compile without -Isrc and fail to find amy.h).
--- a/src/patches.c
+++ b/src/patches.c
@@ -12,6 +12,13 @@
 uint32_t max_num_memory_patches = 0;
 struct delta **memory_patch_deltas = NULL;
 uint16_t *memory_patch_oscs = NULL;
+// Nonzero when the slot's number was invented by patches_store_patch
+// rather than chosen by the caller. Only auto-assigned slots may be
+// recycled (reused on a re-store to the same instrument, freed when
+// that instrument is released): an explicitly numbered patch may be
+// loaded again later by anyone who knows the number, so it is never
+// freed behind the caller's back.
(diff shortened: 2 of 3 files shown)


dpwe:
I'm curious how this came up.  The "memory patch" system was never fully developed and I expected we could get rid of it in light of more direct ways to configure patches.  The comment says 
> every gamma9001 / exported-beat channel is a patch_string instrument

I would expect setting up gamma9001 drum kits to be something like `amy.send(synth=N, patch=384+i)`, which I think bypasses memory patches.  Can you point me to the use of memory patches within gamma9001?

The changes seem fine but I'm more interested in the possibility of removing memory patch.

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1314

PR #1098, opened 2026-08-13
patches: reading back a synth must not deref an unallocated osc

dpwe:
`amy_get_synth_commands()` and `amy_dump_state()` **segfault** on any synth that owns an oscillator nothing ever allocated:

```python
amy.send(synth=1, num_voices=1, oscs_per_voice=1)
amy.get_synth_commands(1)          # SIGSEGV
```

## Why

`synth[]` is an array of *pointers*, NULL until `ensure_osc_allocd()` fills one in, and a voice can legitimately own oscs that were never touched — a synth made with `oscs_per_voice=` and no patch reserves its oscs (`osc_to_voice`, `voice_to_base_osc`) without allocating a single `synthinfo`, and a `patch_string` only allocates the oscs it actually names.

`set_event_for_osc()` read `synth[osc]->max_num_breakpoints` on its first line with no check, and both readers walk every osc of every voice through it.

Measured on `main` before the fix:

| synth config | `get_synth_commands` |
|---|---|
| built-in `patch=0` | ok |
| `patch_string`, 1 osc | ok |
| `patch_string`, 2 oscs | **SEGFAULT** |
| `oscs_per_voice=1` | **SEGFAULT** |
| `oscs_per_voice=4`, 2 voices | **SEGFAULT** |

So not an obscure shape — `oscs_per_voice=N` with no patch is the plain way to make a synth, and it crashed every time. `amy_dump_state()` walks the same path, so state dump was equally dead for those synths.

## The fix

Reading an unallocated osc is not an error: it is at its defaults, so there is nothing to emit for it. Return early instead of dereferencing.

## Test

`tests/test_synth_readout.c` pins all five shapes above plus an undefined synth, wired into `make ctest`. I verified the test actually catches the bug rather than just passing: with the guard reverted it exits **139**.

`make test` 131/131 and the rest of `ctest` unchanged.

Also adds the two ctest binaries to `.gitignore` — `test_patch_slots` has been untracked-but-unignored since #1096 added it.

## How it surfaced

Found while looking at whether a `patch_string`'s stored deltas can be discarded once a synth is configured (they can't today — `patches_load_patch` re-reads them via `instrument_get_patch_number()` whenever `num_voices=` changes). The read-back path is what a "clone voice 0 into the new voices" design would lean on, so it wanted hardening first. That redesign is separate and not in this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,8 @@ tests/test_clock_wrap
 tests/test_sequencer_active
 tests/test_sequencer_bounds
 tests/test_bus_config
+tests/test_patch_slots
+tests/test_synth_readout
 amy-piano
 src/build/
 *.DS_Store*
--- a/Makefile
+++ b/Makefile
@@ -105,7 +105,8 @@ amy-message: $(OBJECTS) src/amy-message.o
 # Plain C tests for things the audio-rendering suite can't reach -- e.g. clock
 # rollovers 50 days out, which you can only hit by fast-forwarding the counters.
 CTESTS = tests/test_clock_wrap tests/test_sequencer_active tests/test_sequencer_bounds \
-         tests/test_bus_config tests/test_patch_slots
+         tests/test_bus_config tests/test_patch_slots \
+         tests/test_synth_readout
 
 # Static pattern rules, so these win over the generic %.o: %.c above (which
 # would compile without -Isrc and fail to find amy.h).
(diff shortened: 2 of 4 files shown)


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1316

PR #1100, opened 2026-08-14
Growing a synth clones its live voice, not its original patch [skip release]

dpwe:
Step 2 of the memory-patch thread (after #1096 and #1098).

## The bug

Changing `num_voices` on a synth that already has voices releases every voice and re-initializes them all, and the only source it had for that was the instrument's **stored patch number**. So:

```python
amy.send(synth=1, patch="v0w0f220Zv1w0f330Z")
amy.send(synth=1, osc=1, wave=amy.SAW_DOWN)   # tweak it
amy.send(synth=1, num_voices=4)               # ...and lose the tweak
```

Growing reverted every voice — voice 0 included, the one that had been playing the edited sound right up until you asked for more of it — to the pristine patch.

A synth with **no** patch couldn't grow at all: `oscs_per_voice=` with no patch leaves the instrument's patch number unset, which fell through to the user-patch branch as 65535, printed `load patch_number 65535 is out of range`, and returned — *after* `patches_voices_for_load_synth` had already reallocated the voices, so the synth was left with voices reserved that the instrument didn't know about.

## The change

Snapshot the live voice into a base-osc-relative delta list *before* the reallocation resets its oscs, and replay that into every voice. `set_event_for_osc()` already knows how to read an osc's config back — it's what the synth readout uses — and it deliberately excludes note and velocity, so a synth cloned mid-note copies its *sound*, not the note being played.

Scope is deliberately narrow:

- An explicitly named `patch_number` or `oscs_per_voice=` loads exactly as before. The clone only stands in where `instrument_get_patch_number()` was the fallback.
- The synth **keeps** whatever patch number it had. #1096's memory-patch slot lifecycle (auto slots recycled on re-store, freed on release) is untouched by this PR.
- The snapshot is released back to the delta pool on every exit path.

## Follow-ups this unblocks

Now that nothing reads the auto-assigned patch number to rebuild a synth, the two remaining pieces can land next: leaving `patch_number` UNSET for auto-assigned memory patches (and freeing the slot as soon as it is loaded), and the `amy_midi.c` bank-number guard that currently computes bank 8 from an allocator artifact.

## Testing

- `tests/test_clone_on_grow.c` (new, wired into `make ctest`): edits survive growth *and* shrink; a hand-configured synth with no patch can grow; an explicit patch load still overrides edits; a ROM-patch synth keeps its edits; 20 grow/shrink cycles leave the delta pool exactly where it started. Verified it **fails 5 checks** with `src/patches.c` reverted, so it catches the bug rather than merely passing.
- `make ctest` — all 7 binaries pass.
- `make test` — 131/131.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,3 +77,4 @@ Mkfile.old
 dkms.conf
 godot/.sconsign.dblite
 godot/src/*.os
+tests/test_clone_on_grow
--- a/Makefile
+++ b/Makefile
@@ -106,7 +106,7 @@ amy-message: $(OBJECTS) src/amy-message.o
 # rollovers 50 days out, which you can only hit by fast-forwarding the counters.
 CTESTS = tests/test_clock_wrap tests/test_sequencer_active tests/test_sequencer_bounds \
          tests/test_bus_config tests/test_patch_slots \
-         tests/test_synth_readout
+         tests/test_synth_readout tests/test_clone_on_grow
 
 # Static pattern rules, so these win over the generic %.o: %.c above (which
 # would compile without -Isrc and fail to find amy.h).
--- a/src/patches.c
+++ b/src/patches.c
@@ -1200,6 +1200,33 @@ void release_voice_oscs(int32_t voice, uint32_t time) {
     }
 }
 
+uint16_t snapshot_voice_config(uint8_t instr_num, struct delta **queue, uint32_t time) {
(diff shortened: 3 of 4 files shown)


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1317

PR #1101, opened 2026-08-14
An auto-assigned patch number is a courier, not a patch

dpwe:
**Stacked on #1100** — base is `clone_on_grow`, so review that one first; this PR's own diff is the second commit. GitHub will retarget it to `main` when #1100 merges.

Steps 2b and 3 of the memory-patch thread, which #1100 unblocked.

## An auto-assigned patch number is a courier, not a patch

A `patch_string` with no `patch_number` gets one invented for it by `patches_store_patch`, purely so the deltas have somewhere to live until `patches_load_patch` reads them. The caller never sees that number and can never ask for it again — but the instrument stored it, and everything downstream then treated an allocator artifact as if it meant something.

Now that growing a synth clones its live voice (#1100), nothing needs the number after the load, so it is consumed there:

- the instrument records **no** patch number (UNSET) rather than one nobody can use;
- the slot goes **back to the pool on the way out**, whether the load worked or not (all exits share one cleanup path).

A slot is now held for exactly the span of one load instead of the lifetime of a synth, so the pool contains only what callers explicitly numbered.

## This deletes the #1096 machinery for auto slots

Both halves become unreachable and go:

- `patches_store_patch` no longer reaches into the addressed instrument to recycle its slot — there is nothing to recycle, the slot was freed at load;
- releasing an instrument no longer hunts for an auto slot to free — an instrument never holds one.

Explicitly numbered patches are untouched throughout: stored, kept on the instrument, never freed behind the caller's back.

## The `amy_midi` bank guard

`amy_received_program_change` masked and shifted whatever `instrument_get_patch_number()` returned. A synth with no patch number computes `(65535 & 0xFF80) >> 7` = **bank 511**, putting a bare program change 65408 patches past anything defined — so the PC did nothing at all. (A synth that doesn't exist reports `-1`, same story.) It now defaults to bank 0 (Juno) in both cases.

Per your call on the last thread, an **explicit** user patch number keeps its bank logic: 1030 is bank 8, and a bare PC there still walks that bank. Only the machine-generated case is guarded.

## Testing

- `tests/test_patch_slots.c` rewritten to the new contract: a config leaves no slot standing, the synth records no patch number, 40 reconfigures burn nothing, 40 configure/release cycles leak nothing. The explicit-number, dry-pool and release-is-quiet tests are unchanged and still pass.
- New program-change test in the same file — it **fails without the `amy_midi` guard** (`PC 3 loaded built-in patch 3 (got 65535)`), and pins that an explicitly numbered patch still resolves its own bank.
- `make ctest` 7/7, `make test` 131/131.
- `tulipcc` (`tulip/macos/build.sh`) builds clean against this commit.

### DIFF
--- a/src/amy_midi.c
+++ b/src/amy_midi.c
@@ -158,7 +158,17 @@ void amy_received_program_change(uint8_t channel, uint8_t program) {
     if (bank_number < 0) {
         // If the bank hasn't been set, stay within the block of 128 of the current patch
         // (so e.g. DX7 voices remain DX7).
-        bank_number = (instrument_get_patch_number(e.synth) & 0xFF80) >> 7;
+        int synth_patch = instrument_get_patch_number(e.synth);
+        // ...but only if there IS a current patch. A synth built from a
+        // patch_string carries no patch number (the auto-assigned one is an
+        // allocator artifact, released as soon as it is loaded), and a synth
+        // that doesn't exist reports -1. Both used to be masked and shifted
+        // into a nonsense bank -- 511 for the unset sentinel -- which put the
+        // program change hundreds of patches past anything defined, so a bare
+        // PC on such a channel did nothing at all. Default to bank 0 (Juno).
+        uint16_t synth_patch_u16 = (uint16_t)synth_patch;
+        if (synth_patch < 0 || AMY_IS_UNSET(synth_patch_u16))  synth_patch = 0;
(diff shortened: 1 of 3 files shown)


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1318

PR #1108, opened 2026-08-16
log2_lut: floor the argument instead of spinning on it

dpwe:
## The hang

`log2_lut()` normalizes its argument into `[1, 2)` by shifting, so a non-positive one never returns — shifting zero left leaves zero. The precondition was a commented-out `// assert(x > 0);`, and two callers didn't meet it:

- **`map_60dB_to_01f`** guarded `lin == 0`, in *float*. That misses a negative amplitude, a NaN, and — in fixed-point builds, 23 fractional bits — any amplitude too small to survive `F2S`. Any of those wedges the render thread: no crash, no message, just silence and a pegged core.
- **The DX7 attack branch** in `envelope.c` feeds it `MAP_ATTACK_LEVEL_S`, which reaches zero once a breakpoint value passes ~2.37. A hand-written patch can ask for that; no fuzzing needed.

The other three call sites already guarantee a positive argument (`logfreq_of_freq` has `if (freq<=0) return 0;`, the envelope decay branch floors with `MAX(F2S(BREAKPOINT_EPS), …)`).

Backtrace from the wedged process:

```
#0  log2_lut (x=0) at src/log2_exp2.c:17     // while (x < F2S(1.0f))
#1  map_60dB_to_01f (lin=<optimized out>) at src/amy.c:660
#3  amp_combine_controls (...) at src/amy.c:1817
#4  hold_and_modify (osc=0) at src/amy.c:1886
#5  render_osc_wave (...) / amy_render / amy_simple_fill_buffer
```

## The fix, in three places deliberately

- **`log2_lut` floors its argument**, branch-free (`MAX`, which takes the floor for NaN too since `NaN > y` is false). This is the hot path, so it was measured — see below.
- **`map_60dB_to_01f` keeps its own guard**, now covering everything that isn't safely positive, so silence still returns the `-10` sentinel rather than a very negative number. Same single comparison it always had.
- **`envelope.c` floors the mapped attack levels** the way the decay branch below it already floors `v0`/`v1`. The `log2_lut` guard alone would leave both levels equal there and divide by zero computing `t_const`.

## Cost, measured on AMYboard

`tools/arduino_loadsweep`, 6-note chord, 20 s captures:

| | runs (final) |
|---|---|
| main | 2514 / 2515 / 2556 |
| this PR | 2520 / 2520 / 2521 |

**+5 to +6, about +0.2%**, and +284 bytes of flash. The rig's own run-to-run spread (that 2556, max 2841) is larger than the effect.

## How it was found, and why CI never caught it

`TestFuzzWireParser` hangs at its 92nd random message on x86 — an amp coefficient lands on the wrong side of the underflow there, while the same message on arm64 does not. That test arrived in #1089, whose CI run was **cancelled**, and no C/C++ CI run has completed on main since. So this has been sitting in main unnoticed, and it is what has been stalling every recent PR's Python jobs at 'in progress' for hours.

## Testing

- New `tests/test_log2_lut.c`: zero, negative, NaN, underflow, the DX7 ceiling, and that the documented mapping still holds (`1.0→1`, `0.001→0`, round-trip). It calls `alarm(30)`, so a missing guard fails in half a minute rather than leaving a job 'in progress' for six hours. **Verified it hangs (exit 142) with these changes reverted.**
- `make ctest` 8/8, `make test` 132/132 locally (arm64, where the fuzz test passed before too — x86 CI on this PR is the real check).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,4 @@ dkms.conf
 godot/.sconsign.dblite
 godot/src/*.os
 tests/test_clone_on_grow
+tests/test_log2_lut
--- a/Makefile
+++ b/Makefile
@@ -106,7 +106,7 @@ amy-message: $(OBJECTS) src/amy-message.o
 # rollovers 50 days out, which you can only hit by fast-forwarding the counters.
 CTESTS = tests/test_clock_wrap tests/test_sequencer_active tests/test_sequencer_bounds \
          tests/test_bus_config tests/test_patch_slots \
-         tests/test_synth_readout tests/test_clone_on_grow
+         tests/test_synth_readout tests/test_log2_lut tests/test_clone_on_grow
 
 # Static pattern rules, so these win over the generic %.o: %.c above (which
 # would compile without -Isrc and fail to find amy.h).
--- a/src/amy.c
+++ b/src/amy.c
@@ -652,9 +652,18 @@ void add_delta_to_queue(struct delta *d, struct delta **queue) {
 
 }
 
+// The smallest amplitude this maps rather than treating as silence. log2_lut
+// requires a strictly positive argument -- it normalizes by shifting, so zero
+// shifts to zero forever -- and `lin == 0` did not cover everything that
+// reaches it as zero or worse: a negative amplitude, a NaN, or (in fixed-point
+// builds, 23 fractional bits) any amplitude that underflows the conversion.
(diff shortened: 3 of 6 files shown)


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1322

PR #1116, opened 2026-08-17
Add a per-osc distortion stage: soft clip, wavefolder, bitcrusher ('G' sub-commands)

rt-rtos:
Adds an optional waveshaping stage to each oscillator, and to a voice's SILENT
head osc to shape the summed voice - built for my AMY groovebox, and in
use/testing for a couple of days.
Applied after rendering and before the filter, so drive pushes harmonics out of simple waves and the
filter (or its envelope sweep) shapes the result. Three types:

- `DIST_CLIP` - drive into a cubic soft knee (`y = v - v^3/3`, clamped). Unity
  small-signal gain, so drive is the only gain in the path.
- `DIST_FOLD` - triangle-map wavefolder: `1 - |((drive*x + 1) mod 4) - 2|`.
  Identity on [-1, 1], reflects beyond, with no data-dependent reflect loop.
- `DIST_CRUSH` - bit-depth quantizer plus a sample-and-hold rate reducer, both
  applied at the hold refresh so the two compose.

Params ride `G` sub-commands: a digit after `G` selects `filter_type` as
before, a letter selects a distortion command - `GC1`/`GF1` enable clip and
fold (0 turns the stage off), `GH<bits>,<rate>` enables the bitcrusher (`GH0`
off), `GD<drive>` and `GM<mix>` set the drive and wet/dry shared by every
type - with matching `amy_event` fields and delta params. Clamping happens once at delta apply, so the per-sample loops are
check-free and call-free. No allocation: CRUSH's state is two words in
synthinfo. With the stage unset or off the render path is untouched.

## Design notes

**Per-osc, not per-bus.** On a bus the nonlinearity sees the mix and produces
intermodulation between voices; per-osc it is true per-note waveshaping, and it
costs nothing on oscs that do not enable it.

**`dist_process` returns its own output max.** FOLD amplifies quiet input - a
release tail at drive 8 maps up to near full scale - so reusing the
pre-distortion max in the `OSC_IN_RELEASE && max_val < AMP_THRESH` check
terminates oscs that are still audible.

**The pre-gain uses `MUL6A_SS`, not `MUL4_SS`.** `MUL4_SS` is valid only for
results in [-16, 16), and `drive * x` leaves that range whenever
`drive * |x| >= 16`; the int32 product then wraps sign, turning a full-scale
peak into a full-scale trough. `MUL6A_SS` is the documented `(sample, scale)`
form, reaches [-64, 64), and shifts the sample argument two bits less, so the
clean path gains resolution: -95.2 vs -89.1 dBFS RMS error on a unity-drive
passthrough.

**CRUSH rounds to nearest rather than truncating.** Truncation is a DC offset
of half a step (-0.5 FS at 1 bit), identical in every voice, so it sums
linearly across polyphony and the echo integrates it -
`delay_line_in_out_fixed_delay()` has DC gain `1/(1 - feedback)`, with the
damping branch passing DC at unity. Past full scale the master soft-clip pins
the whole mix to a rail. Rounding costs one add and measures 0 DC at every
depth. It also repairs the release tail: values below half a step round to
silence, so a 1-bit note decays away instead of holding a constant-amplitude
square until `amp_combine_controls` hard-zeros it at `AMP_THRESH_PLUS`.

**Known limitations, by design.** No oversampling - FOLD at high drive and
CRUSH rate reduction alias audibly. The amp envelope
is already baked into the osc buffer when the stage runs, so distortion depth
follows the envelope: notes clean up as they decay and velocity changes timbre.
`bits >= 24` is the true quantizer bypass (`SAMPLE` has 23 fraction bits).

## Performance

ESP32-S3 at 240 MHz, 48 kHz, fixed-point, LTO, 256-sample blocks (1,280,000
cycles per block of budget). Workload: 8 sustained saws, scenes identical
apart from the `C` message, per-block cycles from an on-target headless
harness, median of 3 boots.

| Type | cyc/sample/osc | % of block budget per osc |
|---|---|---|
| CLIP | 28.5 | 0.57% |
| FOLD | 24.0 | 0.48% |
| CRUSH, rate 8 | 17.8 | 0.36% |
| CRUSH, rate 1 | 21.0 | 0.42% |

Eight simultaneously distorted oscs cost 4.7 percentage points of block
headroom. With the stage off the cost is one `dist_type` compare per osc per
block.

Parameters other than rate sit at or under the 0.98% measurement floor
(established by a control scene byte-identical to the baseline): drive spans
0.53% across 1/4/16 on CLIP and 0.92% on FOLD, bit depth 1.16% and
non-monotonic, 50% vs 100% mix 0.61%. Rate is the exception because the
quantizer sits inside `if (count == 0)` while the rest of the loop is paid
every sample: cost fits `18.58 + 3.62/rate` cycles per sample (rates 1, 8, 64
within 0.02), so rate 1 costs 2.58% more than rate 8 while 8 vs 64 differs
only 0.34%. The amortized refill is also why CRUSH undercuts CLIP's cubic and
FOLD's triangle map, which run unconditionally.

Under LTO `dist_process` inlines into `render_osc_wave` and all four loops ship
as Xtensa hardware zero-overhead loops (`loop`, trip count 0x100), CRUSH's
internal refill branch included.

## Verification

Host build, fixed-point. CLIP raises a full-scale sine's crest factor
0.709 -> 0.979 with the peak saturated below the rails. FOLD is transparent at
drive 1 and its HF energy proxy grows monotonically with drive, 0.031 -> 0.314
over drive 1..8. CRUSH at 4 bits yields 16 quantization levels(IIRC) and rate 8 holds
 8-sample plateaus. Params fan out through instrument voices. With the
stage unset, output is bit-identical to master over sine, Juno and DX7 renders
(FNV-1a over 60 blocks). Also running on ESP32-S3 hardware at 48 kHz.

### Spectrograms

Captured off the ESP32-S3 over USB audio, log frequency axis. All four plots
share the same source: a repeating D2 sine at 108 BPM, fully wet (mix 1), so
the differences between plots are the distortion stage alone.
(except vid)

**CLIP - harmonic stack grows with drive.**

clip

Nearly clean on the left; as drive steps up, each note carries a taller stack
of discrete, evenly spaced partials - a harmonic series, not broadband noise.
Within each note the stack rises and falls with the amp envelope: the
level-dependence of the waveshaper.

**FOLD - full-band comb from the first fold.**

fold

Same sweep, far denser: once the signal crosses the first fold point the
comb jumps to high order rather than filling in gradually, where CLIP at
comparable drive is still building its low harmonics. The per-note fan shapes
are the fold count tracking the envelope - more folds at the peak, fewer as
the note decays.

**CRUSH rate reduction, then drive - aliasing lattice.**

downsamplethendrive

First the sample-and-hold rate steps up on the plain sine: the hold's alias
images of the harmonics march downward as the effective sample rate falls -
inharmonic content moving with `rate`, not pitch. Then drive is added on top
of the downsampled signal and the plot snaps into a full-band lattice: the
saturated staircase generates harmonics of every alias at once.

**CRUSH bit depth - the noise floor climbs ~6 dB per bit.**

depth

Bits step from high to low. The left half is visually and audibly clean -
quantization at high depths sits below the floor. As bits drop, each note
throws a harmonic ladder reaching further up the band, and within a note the
products sweep as the decaying envelope slides the signal across quantization
steps. The low band stays at constant level throughout.

A short clip with audio shows a different sweep with the waveform alongside -
the sine visibly ladder-stepping as depth falls, inaudible turning into
audible content:

https://github.com/user-attachments/assets/c46f7634-5b52-4ae1-b07f-82622da5f0a9

## Notes

The wire encoding follows the `G` sub-command shape proposed in review; the
`dist_*` field naming and the 72..76 param ids are still open. Everything else follows the chorus/echo idiom, and
serialization parity is done: `sprint_event`, `event_addresses_oscs`, and
delta-to-event readback all carry the new fields.

I am open to reworking the particulars if needed upon review.

---

#### Update:  

this PR is deliberately just the base primitive - the shaper applied per-oscillator (pre-filter, inside the envelope). dist_block is written summing-scope-agnostic on purpose, so the same three types can later run at different scopes. I've held that full four-scope build-out back rather than pile it into one PR - it's only worth doing if the primitive is wanted here. Per-voice (the SILENT head) is now in this PR; coef modulation stays
prototyped on top. 

**Here´s a diagram sketch of what I´m thinking:**

```mermaid
---
config:
  layout: elk
  look: classic
  theme: dark
---
flowchart
  subgraph CTRL["CONTROL RATE - once per osc per block, 187.5 Hz at 48 kHz"]
    LFO["mod source osc<br/>role = SYNTH_IS_MOD_SOURCE<br/>mod_value refreshed once per block"]
    SRC["other coef sources<br/>CONST, NOTE, VEL, EG0, EG1, BEND, EXT0, EXT1"]
    CI["hold_and_modify: ctrl_inputs[NUM_COMBO_COEFS]"]
    CMB["combine_controls<br/>dist_logdrive_coefs -> exp2f -> drive<br/>dist_mix_coefs -> clamp -> mix"]
    MS["msynth: dist_drive, dist_mix<br/>read once, above the sample loop"]
    LFO -->|COEF_MOD0 / MOD1| CI
    SRC --> CI
    CI --> CMB
    CMB --> MS
  end

  OSC["oscillator"]

  subgraph VOICE["MODULATED DOMAIN - per-note, coef rail reaches every stage"]
    GEN["osc generator<br/>render_saw / render_algo / render_ks"]
    DP["<b>PER-OSC SCOPE</b> #1116 - dist_process<br/>post-envelope, pre-filter"]
    FLT["filter_process"]
    ACC["per_osc_fb: chained-osc accumulation<br/>= ONE VOICE"]
    SIL["SILENT head osc:<br/>render_envelope + filter_process"]
    GEN --> DP
    DP --> FLT
    FLT --> ACC
    ACC --> SIL
  end

  BOUND["<b>DOMAIN BOUNDARY - mix_with_pan into fbl core bus</b><br/>ctrl coef -> per bus/global"]

  subgraph POST["STATIC POST-MIX DOMAIN - set by delta, no coef rail"]
    direction TB
    T3(["<b>PER-BUS SCOPE</b> planned - tap<br/>no stage; hook only at chain end"])
    FX["per-bus FX: EQ -> chorus -> echo -> reverb<br/>every voice on the bus"]
    SUM["sum buses x volume_scale"]
    T4(["<b>GLOBAL SCOPE</b> planned - tap<br/>the entire mix"])
    CLIP["soft clipping -> int16"]
    T3 --> FX
    FX --> SUM
    SUM --> T4
    T4 --> CLIP
  end

  OSC --> GEN
  SIL --> BOUND
  BOUND --> T3
  SIL -.-> PER["<b>PER-VOICE SCOPE</b> in this PR - SILENT head osc<br/>dist_block on the summed chain, post-envelope, pre-filter"]
  MS -.->|"per-osc coefs"| DP
  MS -.->|"head osc's own coefs:<br/>its EG1 and LFO shape the whole voice"| PER
  CLIP ==> OUT["output"]
```


rt-rtos:
Two follow-ons from building this into my synth's editor. Both separable -
happy to leave them for later or fold either in.

`dist_process` runs only in the `wave != SILENT` branch, before chained oscs
are summed, so a SILENT control osc applies env and filter but not distortion.
Adding it there gives a per-voice summing node - `chained_osc` is
base-osc-relative, so each voice has its own head, buffer and state. It is
musically distinct from per-osc, since the nonlinearity is what generates the
intermodulation products, so it matters whether the oscs sit inside the shaper
together. Host sim, Juno patch 0: rms 755.6 -> 2069.0 shaping the head, 1019.0
shaping the chain's sounding members instead. Distortion-off and ALGO-head
cases stay byte-identical.

That needs `dist_process` split into `dist_block(block, len, cfg, st)` over an
explicit `dist_config_t` and `dist_state_t`, with the per-osc wrapper and
`amy_event`'s flat wire fields unchanged.

Potential surfaces/scopes: per-osc (this PR), per-voice silent-head (least invasive)
or new voice level mix tap, and per-bus - which needs nothing in AMY, since
`amy_external_bus_postprocess_hook` already fires after the FX chain and
before mixdown.

---

I am unsure of what the standard approach to modulation is in cases like this.

Per-osc/wave waveshaper feels like it belongs on ctrl coef rail, but there are some
level coupling concerns.

dpwe:
Do you have any examples of how this works musically?

I feel like the nonlinear distortion (soft-clip) wants to be post-envelope (so you get dynamic harmonic content in response to envelope variation), but right now envelopes are usually applied in the SILENT osc.  I think having an envelope -> distortion -> filter sequence (which maybe is only engaged in the head SILENT osc) seems promising.

dpwe:
We could overload, say, `G` (`filter_type`) for a new set of commands to control these modes individually e.g. `GC1` to enable clip, `GF1` to enable fold, `GHbits,rate` for crush.  Are `drive` and `mix` shared across all effects?  With separate commands you could imagine combining them, right?  Or does that not make any sense?

rt-rtos:
Quick answers now, proper reply with audio + spectrograms when I'm back at the hardware:

It's already post-envelope. Amp eg applies inside the wave render, and this stage runs post-render, pre-filter - so the chain is exactly your env → dist → filter. Hard hits bark, tails clean up as the EG decays.

SILENT-head distortion + coef modulation exists in a follow-up. Can add both or port only the silent head to this PR. (shaper runs once per voice on the summed chain, post-envelope, pre-filter).

Drive/mix are shared (drive = clip pre-gain / fold depth; crush = pre-gain into saturator, but bits/rate is the main knob instead; mix = wet/dry for all). One type per osc, no stacking as is but I have thought about it and would be happy to prototype.

 C encoding vs G overload; only constraint is filter + dist coexist on an osc, and drive/mix want to stay modulation-addressable.

More detail + audio captures to follow later today. 

---

Edit: Update on stacking: it's open, and mechanically cheap - the shaper takes caller-owned config/state, so a chain is just extra passes (cost additive, ~70 cycles/sample/osc with all three). 
The real work is per-stage drive and a defined order. Your G encoding actually carries this well: dispatch on the char after G (digit → filter_type as today, letter → dist stage) keeps back-compat, lets filter and dist coexist, and makes stages independently toggleable (GC0) - which a single type field can't. 
Only hard requirement either way: drive/mix stay modulation-addressable. 
I'll include a crush→clip A/B (one osc vs osc+head)in the audio examples.

rt-rtos:
Pushed two commits, and rebased the branch onto 1.2.154 while at it.

**`G` encoding adopted.** A digit after `G` is `filter_type` as before; a
letter dispatches to the distortion stage, the same pattern as the `i`
synth-layer sub-commands. `GC1`/`GF1` enable clip and fold, 0 turns the stage
off; `GH<bits>,<rate>` enables the crusher (`GH0` off); `GD<drive>` and
`GM<mix>` carry the shared drive and wet/dry. `C` is freed.

Two properties I wanted from this split, beyond freeing the letter:

- per-stage enables mean stacking would need no further wire change - a stage
  per letter, with per-stage order and state as the actual work;
- `GD`/`GM` can grow from scalars into control-coef vectors under the same
  letters if drive/mix modulation lands: `parse_coef_message` on a single
  value sets just the constant term, so a coef list there is a pure extension.

Python mirrors the wire: `dist_clip`/`dist_fold`/`dist_crush`/`dist_drive`/
`dist_mix` kwargs replace the `dist` list.

**SILENT-head distortion is in.** `dist_process` is split into
`dist_block(block, len, cfg, st)` over explicit `dist_config_t`/`dist_state_t`,
and a SILENT chained-osc head with a distortion type set runs it once on the
voice's summed chain - after the envelope, before the head's filter, so it is
the `envelope -> distortion -> filter` order per voice you described. Per-osc
and head compose: e.g. crush per osc, clip on the head as glue.

Verification: parse/print round-trip over the new commands (enables alone,
full messages, `filter_type` and distortion coexisting in both orders, off
forms); head clip on a chained saw voice lifts RMS 1174 -> 1530 at drive 8,
enable-then-disable renders identical to dry; with the stage unset, output is
unchanged.

PR body is updated to match. Audio + spectrograms for the per-osc vs
SILENT-head A/B still to come.

rt-rtos:
**Volume warning** - start with speakers or headphones at a safe level and adjust upwards just to be safe, as these will blast at full volume if being unmuted while playing.

---

Audio demos, as promised. Everything is rendered through the Python
module built from this branch (`make amy-module`), so each clip is
reproducible from its `amy.send()` sequence - the full render script is
collapsed at the bottom. Every clip is peak-normalized to -6 dBFS so louder
doesn´t read as "better" in an A/B, and every wet clip has a `-dry` twin with
identical notes and timing, distortion stripped. The videos are log-frequency
spectrograms with the clip's audio.

**1 - CLIP is post-envelope.** One sine patch (`dist_clip`, drive 3, full
wet), the same note played hard (vel 1.0) then soft (vel 0.2). The hard hit
barks and audibly cleans up as the EG walks the signal back down the transfer
curve; the soft hit never reaches the knee. This is the
`envelope -> distortion -> filter` order doing its job per note.


https://github.com/user-attachments/assets/59a768f8-1ac5-4168-a971-9e3a27cd6591



<details><summary>dry twin</summary>


https://github.com/user-attachments/assets/c3c1c707-3bf4-42cc-b61f-a4e78a66a72e



</details>

**2 - FOLD drive sweep.** Sustained sine, drive swept 0 to 8. Harmonic bloom
from a pure wave, no filter involved - the spectrogram shows the partial fan
unfolding as drive rises.



https://github.com/user-attachments/assets/0b015871-ce23-4b9c-81c5-4fb83447f902


<details><summary>dry twin</summary>


https://github.com/user-attachments/assets/4e787b4a-cd09-4f02-8977-e8440973c0c5



</details>

**3 - FOLD pluck.** Triangle with a fast-decay EG into a fixed fold (drive 5).
Same envelope interaction as demo 1, opposite direction of travel: each note
attacks folded and relaxes toward a pure tone inside its own decay - every
note ends purer than it started.


https://github.com/user-attachments/assets/464f0861-39e7-4ccf-933d-d889d697e667



<details><summary>dry twin</summary>


https://github.com/user-attachments/assets/7064ef5a-83d8-4b84-aaa2-444f2c1fdcfd



</details>

**4 - CRUSH, rate only.** Two-octave line under a sample-hold at rate 8, bits
off. The alias partials move against the melody - lo-fi that tracks the notes
rather than sitting at a fixed fizz.


https://github.com/user-attachments/assets/f074844b-b636-4291-8fcb-1bbaf3103313



<details><summary>dry twin</summary>


https://github.com/user-attachments/assets/9f53cd6a-2da1-46f6-8cfb-0ab9cfee939c



</details>

**5 - CRUSH, bits only.** A sine with a 1 s release, one note per bit depth,
stepping 8 down to 1 (rate 1, drive 1). Because the envelope scales the
signal before the stage, the release decays through ever fewer quantization
levels and finally gates out - by 2 bits the tail is gone mid-release, and at
1 bit the note is a short gated burst. That dissolving tail is the audible
signature of bit depth; a loud static source hides it.


https://github.com/user-attachments/assets/5f0e8403-b7c4-420a-a105-4800c3f96c31



Holding bits at 2 and stepping drive 1, 2, 4, 8, 16 shows the interaction:
drive lifts the decaying tail back over the quantization steps (the notes
grow longer again, as stepped grit) and squares the sustain.


https://github.com/user-attachments/assets/671b1ec0-5fe6-4579-9517-06ed513a9668



<details><summary>dry twin (one serves both - identical clean sine line)</summary>


https://github.com/user-attachments/assets/d843601c-6bbc-4e6c-a3b8-b7377d52e56e



</details>

**6 - per-osc vs SILENT-head clip on a chord.** An ET A-major triad
(220 / 277.18 / 329.63 Hz, sines), same drive, rendered twice: clip on each
osc independently, then the triad chained into a SILENT head that clips the
sum once. Per-osc keeps sparse, separate harmonic stacks per note; the head
fills the gaps with sum/difference intermodulation and a low difference-tone
band - the power-chord growl. The per-osc variant uses three independent
oscs deliberately: chain members render into the head's buffer, so per-osc
distortion on a member would shape the accumulated sum, not its own wave.



https://github.com/user-attachments/assets/ebcb7caf-71f1-43fc-9904-179aa017250a




https://github.com/user-attachments/assets/0e2ea972-e957-4928-8228-4224cd815b0f


<details><summary>dry chord</summary>


https://github.com/user-attachments/assets/774e6a26-97e7-4f55-b44e-b93e77054be6



</details>

**7 - 808 kick, drive stepping.** The BD1 ROM preset through CLIP, drive
1 to 16 across six hits - saturation compressing the pitch sweep into a
harder, longer-sustaining kick.


https://github.com/user-attachments/assets/8d52fca9-05fc-4d05-a8fb-512508a6584f



<details><summary>dry twin</summary>


https://github.com/user-attachments/assets/51dada84-b854-4ec6-b24f-904791ddf17d



</details>

**8 - two-bar 808 pattern.** Kick / snare / hats / clap, once through CRUSH
(6 bits, rate 5) on every drum osc, once through CLIP (drive 5). The per-osc
scope is what makes this usable as a kit effect: each drum is shaped against
its own level, not a bus mix.






https://github.com/user-attachments/assets/cce8013c-0a24-4cb2-9763-407a63362010




https://github.com/user-attachments/assets/6d957f38-ce9c-4da6-a475-d85a4cb654f4



<details><summary>dry pattern</summary>


https://github.com/user-attachments/assets/ed9a9af5-8f40-4ea6-9c67-133656c46aab



</details>

**9 - ordering: clip before vs after a resonant filter.** Why the stage sits
pre-filter, made audible. A slow bass line, LPF24 at resonance 4 with an EG
sweep, clip at drive 8, both orders from the same engine: the native per-osc
order (dist -> filter) keeps the resonance a clean whistle over the clipped
saw, with the filter still taming everything above the cutoff. Moving the
filter to a chain member and the clip to its SILENT head reverses the order -
the head shapes the member's filtered output, so the clipper regenerates
harmonics from the resonant peak itself and the spectrum fills to the top:
the "303 into a distortion pedal" sound. Both orders are expressible; the
default keeps the filter in charge.


https://github.com/user-attachments/assets/00118699-6d68-4f71-bcce-1034e7417730





https://github.com/user-attachments/assets/94364b18-ee81-4f89-afea-87f1b1e11f32



<details><summary>dry (filter only, no clip)</summary>



https://github.com/user-attachments/assets/20a26d19-ed97-4832-bc5a-b16d2f429708



</details>

<details>
<summary>demos.py - the exact render script for every clip above</summary>

```python
#!/usr/bin/env python3
"""Audio demos for shorepine/amy PR #1116 (per-osc + SILENT-head distortion).

Renders each demo offline through the amy module (no audio device needed),
writes a WAV, a log-frequency spectrogram PNG, and an mp4 (spectrogram +
audio) per clip.  Every wet clip gets a dry twin with identical notes and
timing so A/B comparisons are exact.

Run from the PR branch checkout after `make amy-module`:
    python demos.py [outdir]
"""
import os
import sys

import numpy as np
import soundfile as sf

import amy

SR = amy.AMY_SAMPLE_RATE
OUTDIR = sys.argv[1] if len(sys.argv) > 1 else os.path.dirname(os.path.abspath(__file__))

# TR-808 ROM presets baked into the Gamma9001 builds (src/pcm_gamma808.h).
BD1, CLAP, HAT_C, HAT_O, SNARE = 0, 3, 9, 10, 12

DIST_KEYS = ('dist_clip', 'dist_fold', 'dist_crush', 'dist_drive', 'dist_mix')


def render_events(events, seconds):
    """events = [(t_sec, kwargs), ...]; render `seconds` of audio, sending
    each event once the clock reaches its time (block-quantized)."""
    amy.restart()
    amy.render(0.05)  # let restart settle
    out = []
    t = 0.0
    for ev_t, kw in sorted(events, key=lambda e: e[0]):
        if ev_t > t:
            out.append(amy.render(ev_t - t))
            t = ev_t
        amy.send(**kw)
    if seconds > t:
        out.append(amy.render(seconds - t))
    return np.vstack(out)[:, 0]  # mono (all demos are center-panned)


def dry_twin(events):
    """Same events with every distortion kwarg removed."""
    out = []
    for t, kw in events:
        kw = {k: v for k, v in kw.items() if k not in DIST_KEYS}
        if kw:
            out.append((t, kw))
    return out


def spectrogram_png(x, path, title):
    import matplotlib
    matplotlib.use('Agg')
    import matplotlib.pyplot as plt
    nfft, hop = 4096, 512
    win = np.hanning(nfft)
    frames = np.lib.stride_tricks.sliding_window_view(
        np.pad(x, (0, nfft)), nfft)[::hop]
    S = np.abs(np.fft.rfft(frames * win, axis=1)).T
    db = 20 * np.log10(S + 1e-7)
    tt = np.arange(S.shape[1]) * hop / SR
    ff = np.fft.rfftfreq(nfft, 1 / SR)
    plt.style.use('dark_background')
    fig, ax = plt.subplots(figsize=(10, 4), dpi=100)
    ax.pcolormesh(tt, ff[1:], db[1:], cmap='magma',
                  vmin=db.max() - 90, vmax=db.max(), shading='auto')
    ax.set_yscale('log')
    ax.set_ylim(30, SR / 2)
    ax.set_xlabel('s')
    ax.set_ylabel('Hz')
    ax.set_title(title, fontsize=10)
    fig.tight_layout()
    fig.savefig(path)
    plt.close(fig)


def emit(name, events, seconds, with_dry=True):
    clips = [(name, events)]
    if with_dry:
        clips.append((name + '-dry', dry_twin(events)))
    for cname, evs in clips:
        x = render_events(evs, seconds)
        peak = np.abs(x).max()
        peak_db = 20 * np.log10(peak + 1e-9)
        rms = np.sqrt((x ** 2).mean())
        # Normalize every clip to -6 dBFS peak so louder never reads as
        # "better" in an A/B; the raw peak is reported alongside.
        x = x * (0.5 / max(peak, 1e-9))
        wav = os.path.join(OUTDIR, cname + '.wav')
        png = os.path.join(OUTDIR, cname + '.png')
        sf.write(wav, (x * 32767).astype(np.int16), SR)
        spectrogram_png(x, png, cname)
        print(f'{cname:26s} {seconds:5.2f}s  peak {peak_db:6.1f} dBFS  rms {rms:.4f}')
        if rms < 1e-4:
            print(f'  WARNING: {cname} is near-silent')


# --- 1. CLIP x envelope: hard hit barks, soft hit stays clean --------------
def demo_clip_envelope():
    patch = dict(osc=0, wave=amy.SINE, bp0='0,1,2500,0.02,300,0',
                 dist_clip=1, dist_drive=3, dist_mix=1)
    ev = [(0.0, patch),
          (0.2, dict(osc=0, note=45, vel=1.0)),
          (2.9, dict(osc=0, vel=0)),
          (3.4, dict(osc=0, note=45, vel=0.2)),
          (6.1, dict(osc=0, vel=0))]
    emit('01-clip-envelope', ev, 6.6)


# --- 2. FOLD drive sweep: harmonic bloom from a pure sine ------------------
def demo_fold_sweep():
    ev = [(0.0, dict(osc=0, wave=amy.SINE, dist_fold=1, dist_drive=1, dist_mix=1)),
          (0.1, dict(osc=0, note=52, vel=1))]
    t = 0.5
    for i in range(33):                      # drive 0 -> 8 in 0.25 steps
        ev.append((t, dict(osc=0, dist_drive=i / 4)))
        t += 0.18
    ev.append((t + 0.2, dict(osc=0, vel=0)))
    emit('02-fold-sweep', ev, t + 0.7)


# --- 3. FOLD pluck: every note ends purer than it started ------------------
def demo_fold_pluck():
    ev = [(0.0, dict(osc=0, wave=amy.TRIANGLE, bp0='0,1,900,0,0,0',
                     dist_fold=1, dist_drive=5, dist_mix=1))]
    t = 0.2
    for n in (48, 52, 55, 60):
        ev.append((t, dict(osc=0, note=n, vel=1)))
        t += 0.7
    emit('03-fold-pluck', ev, t + 0.6)


# --- 4. CRUSH rate: aliasing that tracks the melody ------------------------
def demo_crush_rate():
    ev = [(0.0, dict(osc=0, wave=amy.SAW_DOWN, bp0='0,1,300,0,0,0',
                     dist_crush=[24, 8], dist_mix=1))]
    t = 0.2
    for n in (36, 43, 48, 55, 60, 55, 48, 43):
        ev.append((t, dict(osc=0, note=n, vel=1)))
        t += 0.25
    emit('04-crush-rate', ev, t + 0.5)


# --- 5. CRUSH bits: a decaying sine tail falls through the steps -----------
# The envelope scales the osc before the dist stage, so a 1 s release decays
# through ever fewer quantization levels until it gates out entirely - the
# audible signature of bit depth that a static loud source hides.  Rate 1 =
# no sample-and-hold, so 5a is pure depth; 5b holds bits at 2 and raises
# drive, which lifts the tail back over the steps (and squares the sustain).
CRUSH_NOTE = dict(osc=0, wave=amy.SINE, bp0='5,1,150,0.5,1000,0')

def crush_ladder(name, steps):
    ev = [(0.0, dict(dist_mix=1, **CRUSH_NOTE))]
    t = 0.2
    for kw in steps:
        ev.append((t, dict(osc=0, note=57, vel=1, **kw)))
        ev.append((t + 0.45, dict(osc=0, vel=0)))
        t += 1.6
    emit(name, ev, t)

def demo_crush_bits():
    crush_ladder('05a-crush-bits',
                 [dict(dist_crush=[b, 1], dist_drive=1)
                  for b in (8, 7, 6, 5, 4, 3, 2, 1)])
    crush_ladder('05b-crush-bits-drive',
                 [dict(dist_crush=[2, 1], dist_drive=d)
                  for d in (1, 2, 4, 8, 16)])


# --- 6. Per-osc vs SILENT-head clip on a chord: the intermodulation A/B ----
def chord_events(dist_on):
    """dist_on: 'oscs' | 'head' | None.  ET major triad (intermod lands off
    the harmonic grid).  'oscs' uses three independent oscs so the clip is
    truly per-note: chain members share the head's buffer, so dist on a
    member would process the accumulated sum, not its own wave.  'head'
    chains them into a SILENT head and clips the sum once."""
    triad = (220.0, 277.18, 329.63)
    ev = []
    if dist_on == 'head':
        ev.append((0.0, dict(osc=0, wave=amy.SILENT, chained_osc=1,
                             bp0='0,1,300,1,150,0',
                             dist_clip=1, dist_drive=4, dist_mix=1)))
        for i, f in enumerate(triad):
            kw = dict(osc=1 + i, wave=amy.SINE, freq=f)
            if i < 2:
                kw['chained_osc'] = 2 + i
            ev.append((0.0, kw))
        ev.append((0.2, dict(osc=0, note=60, vel=0.6)))
        ev.append((3.2, dict(osc=0, vel=0)))
    else:
        for i, f in enumerate(triad):
            kw = dict(osc=1 + i, wave=amy.SINE, freq=f, bp0='0,1,300,1,150,0')
            if dist_on == 'oscs':
                kw.update(dist_clip=1, dist_drive=4, dist_mix=1)
            ev.append((0.0, kw))
            ev.append((0.2, dict(osc=1 + i, note=60, vel=0.6)))
            ev.append((3.2, dict(osc=1 + i, vel=0)))
    return ev

def demo_scope_ab():
    emit('06a-clip-per-osc', chord_events('oscs'), 3.8, with_dry=False)
    emit('06b-clip-head', chord_events('head'), 3.8, with_dry=False)
    emit('06-chord-dry', chord_events(None), 3.8, with_dry=False)


# --- 7. Distorted 808: the real TR-808 kick, drive stepping up -------------
def demo_808():
    ev = [(0.0, dict(osc=0, wave=amy.PCM, preset=BD1,
                     dist_clip=1, dist_drive=1, dist_mix=1))]
    t = 0.3
    for drive in (1, 2, 4, 6, 10, 16):
        ev.append((t - 0.05, dict(osc=0, dist_drive=drive)))
        ev.append((t, dict(osc=0, note=48, vel=1)))
        t += 0.9
    emit('07-808-drive', ev, t + 0.3)


# --- 8. Lo-fi 808 pattern: CRUSH on a beat, SP style -----------------------
def beat_events(fx):
    """fx: None | 'crush' | 'clip'.  Two bars of 4/4 at 100 BPM."""
    kits = ((0, BD1), (1, SNARE), (2, HAT_C), (3, HAT_O), (4, CLAP))
    ev = [(0.0, dict(osc=o, wave=amy.PCM, preset=p)) for o, p in kits]
    if fx == 'crush':
        ev += [(0.05, dict(osc=o, dist_crush=[6, 5], dist_mix=1)) for o, _ in kits]
    elif fx == 'clip':
        ev += [(0.05, dict(osc=o, dist_clip=1, dist_drive=5, dist_mix=1)) for o, _ in kits]
    q = 60.0 / 100.0                       # quarter note
    for bar in range(2):
        b = 0.2 + bar * 4 * q
        ev.append((b, dict(osc=0, note=48, vel=1)))                 # kick 1
        ev.append((b + 1.5 * q, dict(osc=0, note=48, vel=0.9)))     # kick 2.5
        ev.append((b + 1 * q, dict(osc=1, note=45, vel=0.8)))       # snare 2
        ev.append((b + 3 * q, dict(osc=1, note=45, vel=0.9)))       # snare 4
        ev.append((b + 3 * q, dict(osc=4, note=60, vel=0.6)))       # clap on 4
        for e in range(8):                                          # 8th hats
            ev.append((b + e * q / 2, dict(osc=2, note=60, vel=0.5 if e % 2 else 0.7)))
        ev.append((b + 3.5 * q, dict(osc=3, note=60, vel=0.6)))     # open hat
    return ev

def demo_beat():
    total = 0.2 + 8 * 0.6 + 0.6
    emit('08a-beat-crush', beat_events('crush'), total, with_dry=False)
    emit('08b-beat-clip', beat_events('clip'), total, with_dry=False)
    emit('08-beat-dry', beat_events(None), total, with_dry=False)


# --- 9. Ordering A/B: clip before vs after the resonant filter -------------
# Per-osc dist runs BEFORE that osc's filter, so the resonance sweep stays a
# clean whistle over the clipped saw (09a).  The other order is reachable by
# construction: give the filter to a chain member and the clip to its SILENT
# head - the head processes the member's filtered output, so the clipper
# regenerates harmonics from the resonant peak itself (09b, the "303 into a
# distortion pedal" sound).  Same notes, same sweep, same drive throughout.
ORDER_FLT = dict(filter_type=amy.FILTER_LPF24, resonance=4,
                 filter_freq={'const': 200, 'eg1': 4}, bp1='0,1,1200,0,0,0')
ORDER_ENV = dict(bp0='0,1,200,1,150,0')
ORDER_DIST = dict(dist_clip=1, dist_drive=8, dist_mix=1)
ORDER_LINE = ((33, 1.5), (31, 1.5), (36, 1.5), (33, 2.2))

def order_events(order):
    """order: 'before' | 'after' | None (dry).  'before' = one osc, dist ->
    filter (the native order).  'after' = filter on the member, clip on the
    SILENT head; note/vel at the head propagate down the chain."""
    if order == 'after':
        ev = [(0.0, dict(osc=0, wave=amy.SILENT, chained_osc=1,
                         **ORDER_ENV, **ORDER_DIST)),
              (0.0, dict(osc=1, wave=amy.SAW_DOWN, **ORDER_FLT))]
    else:
        kw = dict(osc=0, wave=amy.SAW_DOWN, **ORDER_ENV, **ORDER_FLT)
        if order == 'before':
            kw.update(ORDER_DIST)
        ev = [(0.0, kw)]
    t = 0.2
    for n, dur in ORDER_LINE:
        ev.append((t, dict(osc=0, note=n, vel=1)))
        ev.append((t + dur - 0.25, dict(osc=0, vel=0)))
        t += dur
    return ev, t + 0.3

def demo_filter_order():
    for name, order in (('09a-clip-before-filter', 'before'),
                        ('09b-clip-after-filter', 'after'),
                        ('09-filter-order-dry', None)):
        ev, secs = order_events(order)
        emit(name, ev, secs, with_dry=False)


if __name__ == '__main__':
    print(f'AMY {SR} Hz -> {OUTDIR}')
    demo_clip_envelope()
    demo_fold_sweep()
    demo_fold_pluck()
    demo_crush_rate()
    demo_crush_bits()
    demo_scope_ab()
    demo_808()
    demo_beat()
    demo_filter_order()
    print('done')
```

</details>

src/amy.c
```diff
@@ -857,6 +858,11 @@ void amy_event_to_deltas_queue(amy_event *e, uint16_t base_osc, uint16_t oscs_pe
     }
     EVENT_TO_DELTA_I(note_source_channel, NOTE_SOURCE_CHANNEL)
     EVENT_TO_DELTA_I(filter_type, FILTER_TYPE)
+    EVENT_TO_DELTA_F(dist_type, DIST_TYPE)
```
dpwe: Why is dist_type encoded as as an F?  I would expect EVENT_TO_DELTA_I (i.e., store the DIST_TYPE in the int part of the union).

src/amy.c
```diff
@@ -1551,6 +1564,39 @@ void play_delta(struct delta *d) {
     DELTA_TO_SYNTH_F(RATIO, logratio)
     DELTA_TO_SYNTH_F(RESONANCE, resonance)
     DELTA_TO_SYNTH_I(FILTER_TYPE, filter_type)
+    if (d->param == DIST_TYPE) {
+        // Range-check as float before the cast (huge values are UB to cast).
+        // Restart the rate reducer so a type change can't replay a stale held sample.
+        float type = d->data.f;
```
dpwe: With DIST_TYPE as an I, this is just synth[d->osc]->dist.type = d->data.i

src/amy.c
```diff
@@ -1551,6 +1564,39 @@ void play_delta(struct delta *d) {
...
+    }
+    // Clamp here so dist_process doesn't range-check per block.
+    if (d->param == DIST_DRIVE) {
+        float drive = d->data.f;
+        if (drive < 0.0f) drive = 0.0f;
+        if (drive > DIST_MAX_DRIVE) drive = DIST_MAX_DRIVE;
+        synth[d->osc]->dist.drive = drive;
+    }
+    if (d->param == DIST_BITS) {
+        float bits = d->data.f;
```
dpwe: Probably also wants to be an I

src/amy.c
```diff
@@ -1551,6 +1564,39 @@ void play_delta(struct delta *d) {
...
+        synth[d->osc]->dist.bits = bits;
+    }
+    if (d->param == DIST_RATE) {
+        float rate = d->data.f;
+        if (rate < 1.0f) rate = 1.0f;
+        if (rate > 1024.0f) rate = 1024.0f;
+        synth[d->osc]->dist.rate = rate;
+    }
+    if (d->param == DIST_MIX) {
+        float mix = d->data.f;
```
dpwe: I would write a new macro:
```
#define DELTA_TO_SYNTH_F_CLAMPED(FLAG, FIELD, MINVAL, MAXVAL)  \
    if (d->param == FLAG) { synth[d->osc]->FIELD = MAX(MINVAL, MIN(MAXVAL, d->data.f)); };
```
then these become 
```
DELTA_TO_SYNTH_F_CLAMPED(DIST_DRIVE, dist.drive, 0.f, DIST_MAX_DRIVE);
DELTA_TO_SYNTH_F_CLAMPED(DIST_RATE, dist.rate, 1.f, 1024.f);
DELTA_TO_SYNTH_F_CLAMPED(DIST_MIX, dist.mix, 0.f, 1.f);
```
We could probably have DELTA_TO_SYNTH_I_CLAMPED for bits.  We'd still need to special-case DIST_TYPE since you want to reset the state when it's changed.

dpwe (commented):
So it's a bit weird in that `GC1` implicitly turns off `GH` or whatever which makes the state-to-wirecode seem complicated.  I'd like to add macros to keep play_delta short and readable.  Otherwise looks good.

dpwe:
Thanks so much for the examples!  The sounds are gorgeous!  It would be really nice if you could put them in their own docs/distortions.md file so users can understand what these effects can do.

dpwe:
Can you merge with main then rebuild the Godot and JS bindings?  The recent PCM changes are confounding the merge.  We need to keep both sets of changes in amy.h

rt-rtos:
All of it adopted - `dist_type` and `dist_bits` to ints (dist_rate too: the DSP already truncates it to a whole sample count), your `DELTA_TO_SYNTH_F_CLAMPED` as sketched plus an `_I_CLAMPED` twin for bits/rate, `DIST_TYPE` kept special-cased for the state reset. And docs/distortions.md built from the demo material. Commits incoming.

On GC1 implicitly turning off GH: that is the one-type-slot interim showing through. The per-stage letters were chosen so that stacking needs no wire change: with independent enables in the engine, `GC/GF/GH` become true per-stage toggles and state-to-wirecode collapses to 1:1. Prototyped on top of this branch: https://github.com/rt-rtos/amy/tree/proto/dist-stack - a stage bitmask replaces the type slot, enabled stages run as their own passes in a fixed clip -> fold -> crush order (each loop keeps the zero-overhead form, cost additive, ~70 cycles/sample/osc all three), drive/mix stay shared, and multi-stage messages like `GC1GH6,5` round-trip verbatim. Wire and existing patch strings are unchanged. Happy to pull it into this PR instead if you would rather not merge the interim semantics.

Two small additions folding in. A DC blocker on CRUSH's wet path: the sample-and-hold folds partials near multiples of fs/rate down to sub-audio/DC; a 10 Hz one-pole on the wet signal removes it without touching the lo-fi character (much). And the pre-gain moves from MUL6A_SS to SMULR6: a SILENT-head chain sum runs several oscs' worth of full scale, so drive * x could pass the [-64, 64) product range and wrap sign to a click.


**Two more follow-ups, built out the same way:**

- Drive/mix modulation: `GD`/`GM` grow from scalars into control-coef lists
  (`parse_coef_message` on a single value sets just the constant term, so
  `GD8` means what it meant; `GD1,,2` adds two octaves of drive at full
  velocity). Drive rides a log2 rail like freq, mix combines linearly like
  duty; one shared vector serves the whole stage chain. On top of the
  stacking branch: https://github.com/rt-rtos/amy/tree/proto/dist-coef-rail-v2
- Bus-scope distortion (first in the bus FX chain, so echo and reverb take
  tails of the shaped signal; responds to the whole bus level, which is what
  a mixbus saturator is for):
  https://github.com/rt-rtos/amy/tree/proto/dist-bus-stage

All three branches - stacking included - replace their README with a short
tour of the change plus demo clips in the same format as the ones above.

Separate PRs once this lands, or fold them in here - your call. These
aren't drive-by prototypes: I plan to iterate this into a complete,
polished feature. I just don't want to outrun your review capacity, so
tempo and staging are up to you.

One consideration on that choice: except for final touches, the key pieces
are all in place (I think) - if the prototypes survive review as they stand, the
feature is close to complete. The one open scope question is a global-scope
stage, which is more or less a bonus on top of per-bus. Up to you if you find it easier all at once or in pieces - and no need to hurry on my account, I'm in no rush to merge. Thanks.

rt-rtos:
Will resolve, one sec.

---
edit:
done, pushed.

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1335

dpwe:
Looks great. 

I like the next steps (permit stacking of distortions, controlcoef for drive, bus-level application). I'm concerned with the cost of precalculating idle CtrlCoefs but I think that can be fixed separately. I like bus-level distortion, probably wouldn't want it also on final mix (was that what was suggested?). Please move ahead with follow-on PRs!

PR #1119, opened 2026-08-18
esp32: only yield past-overload when rendering actually can't keep up

### LINKED ISSUE #1118: vTaskDelay(1) in i2s.c making low latency impossible
I tried a smaller size DMA buffer and smaller number of samples per cycle on an esp32p4 in order to make the latency lower.

Out of the box on the esp32 p4: about 34ms latency, mainly the DMA buffer.
Wrote some code which flipped a gpio pin and then start a note to see the delay on the scope.

Image

On the scope you can see the final result when I switched to 128 samples per cycle and added:
chan_cfg.dma_desc_num  = 2;
chan_cfg.dma_frame_num = 32;
in esp32_setup_i2s (i2s.c file).

Great, delay is only a few ms, and my performance test is still ok for this chip! Without the problem I describe below amysynth is a remarkable lightweight synth.

However, this only worked when I commented out vTaskDelay(1):
        amy_overload_check(busy_us);
        // If the audio output didn't block at all, we're past overloaded, and this
        // max-priority task would starve everything else on this core (USB, MIDI,
        // the host app).  Audio is already breaking up, so give the rest of the
        // system a tick.
        // if (blocked_us < 150) vTaskDelay(1);
    }

If I keep the overload check as it is, it will destroy sound completely unless using the default 30ms DMA buffer. So somehow this is triggered, even when there isn't any problem at all (if the vTaskDelay is commented out my analysis code doesn't show any problem with overloads), and the vTaskDelay is 10ms, which is so high for my 1.3ms cycle that it keeps creating it's own overloads (...I think...).

The vTaskDelay is meant to give time to other RTOS tasks when the amy synth calculations are finished and just waiting for DMA buffer availability. In my low latency setup that isn't needed, since I'm running amy on both p4 cores and use the LP core to receive messages, there are no other RTOS tasks.

Maybe the solution would be to scale vTaskDelay as 30% of the DMA buffer in stead of always 10ms.

bwhitman:
Fixes #1118.

The `vTaskDelay(1)` added with the overload failsafe (#791) fires on any iteration where the i2s write didn't block (`blocked_us < 150`). That's the right signal with a large DMA ring: past overload the write never blocks, and a max-priority task that never blocks starves USB CDC / MIDI / the host app on its core — the delay is what keeps the board from wedging.

But with a DMA ring shrunk for low latency (as in #1118: 2 descriptors × 32 frames ≈ 1.5 ms against a 2.9 ms block), a perfectly **healthy** just-in-time iteration can also legitimately see `blocked_us ≈ 0`. One tick of delay — 10 ms at a 100 Hz RTOS tick — is then bigger than the entire ring, so a single spurious trigger underruns it; the drained ring makes the *next* write return immediately too, and the delay re-arms on every iteration forever. That's the self-sustaining "it keeps creating its own overloads" behavior reported, and it's why low-latency DMA configurations were unusable without commenting the line out.

The fix gates the yield on genuine overload as well:

```c
if (busy_us >= AMY_BLOCK_US && blocked_us < 150) vTaskDelay(1);
```

- **Starvation protection is fully preserved.** In real overload, rendering one block costs at least the block's real-time budget by definition, so `busy_us >= AMY_BLOCK_US` always holds there — including when the failsafe threshold is disabled (`overload_threshold = 0`), which is why gating on the failsafe's own smoothed counter wouldn't be enough.
- **A healthy loop can never trigger it**, no matter how small the DMA ring or how coarse the tick, because `busy_us` stays under the block budget whenever rendering keeps up. No new config knob, no tick-rate coupling.
- **The death spiral becomes impossible**: a just-in-time write on its own can't start it.

This is the same class of fix as #960 (the MIDI UART timeout that was integer-zero at a 100 Hz tick): tick-rate-dependent behavior replaced by a condition that's only true when it should be.

Simply removing the delay (suggested in the issue) would reintroduce the wedge #791 fixed: the "FreeRTOS already yields when the task blocks" argument is correct for the healthy case, but in genuine overload the task never blocks, so the scheduler never runs anything lower-priority.

### DIFF
--- a/src/i2s.c
+++ b/src/i2s.c
@@ -379,11 +379,18 @@ void esp_fill_audio_buffer_task() {
         // i2s DMA write (or the update-sync wait) above, which is when lower-priority
         // tasks on this core get to run.
         amy_overload_check(busy_us);
-        // If the audio output didn't block at all, we're past overloaded, and this
-        // max-priority task would starve everything else on this core (USB, MIDI,
-        // the host app).  Audio is already breaking up, so give the rest of the
-        // system a tick.
-        if (blocked_us < 150) vTaskDelay(1);
+        // If rendering genuinely can't keep up (a block costs at least its own
+        // real-time budget) AND the audio output didn't block, we're past
+        // overloaded, and this max-priority task would starve everything else
+        // on this core (USB, MIDI, the host app).  Audio is already breaking
+        // up, so give the rest of the system a tick.
+        //
+        // Both conditions matter: with a small DMA ring a healthy just-in-time
+        // iteration can also see blocked_us == 0, and one tick here (10 ms at
+        // a 100 Hz tick rate) can be bigger than the whole ring -- a single
+        // spurious delay underruns it, the drained ring makes the next write
+        // not block either, and the delay re-arms forever (#1118).
+        if (busy_us >= AMY_BLOCK_US && blocked_us < 150) vTaskDelay(1);
     }
 }
 


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1328

PR #1120, opened 2026-08-19
Name the audio devices, and choose the MIDI port on both desktops

bwhitman:
Both halves of "which device?", for an embedding host that has somewhere to put the question -- a menu, a settings pane. amy can already be TOLD which audio device to open and has no way to say what the choices are called; for MIDI it has neither.

## Audio: name the devices, not just print them

`config.playback_device_id` / `capture_device_id` are indices into miniaudio's enumeration, and `amy_print_devices()` writing that list to stderr has been the only way to see it — fine for a command line, no use at all to a menu or a settings pane.

```c
uint32_t amy_audio_device_count(uint8_t dir);
uint32_t amy_audio_device_name(uint8_t dir, uint32_t index, char *buf, uint32_t buflen);
```

Same order the two ids index, so a host can draw the list, take a pick, and store the index straight into the config. The enumeration keeps a context of its own rather than borrowing the device's — the device context is torn down and rebuilt every time the device changes, which is exactly when something is asking what the choices are. Stubbed to "no devices" where audio is not miniaudio (ESP32, Pico, Arduino).

## MIDI: choose the port, in and out

amy has always listened to every CoreMIDI source and sent to every destination. That is the right default — a keyboard plugged in just plays — and is still what -1 means, so nothing about an existing embedding changes.

```c
uint32_t amy_midi_port_count(uint8_t dir);
uint32_t amy_midi_port_name(uint8_t dir, uint32_t index, char *buf, uint32_t buflen);
int32_t  amy_midi_port_selected(uint8_t dir);
void     amy_midi_port_select(uint8_t dir, int32_t index);
```

**The index is resolved once, when it is chosen, into CoreMIDI's persistent unique id**, and everything afterwards asks for that. An index is only true until the next thing is plugged in, and a selection that quietly becomes a *different* device is much worse than one that goes away. `amy_midi_port_selected()` answers `-2` for a port that was chosen and is not on the desk right now, which is deliberately not the same as `-1`.

Two things fell out of it:

- **One input port with every source connected to it**, each tagged with its unique id, rather than a port per source. The per-source arrangement gave the callback nothing to tell them apart by, which is exactly what a chooser needs; filtering in the callback also means a selection change is one store from any thread instead of run-loop work.
- **Sources are reconnected on `kMIDIMsgSetupChanged`.** A keyboard plugged in after startup was a source nothing was ever connected to — it appeared in CoreMIDI and was silent. Invisible while amy took every port; with a list in front of it, it is the first thing anyone tries.

Every other platform gets the four as stubs answering "no ports, no choice", so a host can call them without an `#ifdef`.

## Linux: MIDI at all, and the same chooser

amy had none — `amy_midi.c` said so in a commented-out `fprintf`. This folds in #563 (A. Doinel's contribution), with the four calls above implemented on top so the platform arrives with the chooser rather than needing a second pass. `libasound2-dev` is the one new dependency.

**-1 cannot mean the same thing on the two desktops.** macOS can hold every source and destination open at once, so there -1 is "all of them". ALSA raw MIDI is one handle per device and there is no listening to everything, so here -1 is the **virtual port**: amy appears in the MIDI graph under its own name and anything may connect with `aconnect`. That is the Linux idiom for "no particular device", it is what a soft synth does, and it is what #563 did before there was a choice at all — which is why it stays its own code path, one duplex `snd_rawmidi_open(..., "virtual", ...)`, the exact call anybody has actually run.

A selection is pinned by **name**, since ALSA has no equivalent of CoreMIDI's unique id: `hw:1,0,0` is a position in a list, so plugging a second interface in makes yesterday's string somebody else's device. The string is kept as a hint and the name is what is matched — same guarantee, arrived at differently.

**The read thread owns the handles** and a selection only says what is *wanted*; `snd_rawmidi_read` blocking in one thread while `snd_rawmidi_close` runs in another is the shape that crashes rarely and unreproducibly. It polls the descriptors ALSA hands out with a 100 ms timeout rather than a tight read-and-`usleep` loop — which is also the only reason a selection change or a quit takes effect at all, since a blocking read on an idle keyboard waits forever.

Three things fixed on the way in, each of which would have bitten quietly:

- `midi_out` did not call `midi_out_external_hook`, which `amy_midi.h` requires of every implementation and macOS does — so `amy_config`'s external output hook would never have fired on Linux.
- It wrote to the tx handle with no NULL check, making a send before `run_midi()` a crash rather than a dropped message. macOS guards the same thing.
- A port that would not open called `exit(1)`, taking down the host application over a missing cable. amy is a library.

**Not taken from #563**, and worth saying explicitly: `src/amy-synth.c` no longer links against current amy (`event_generator_for_patch_number` and `print_event` are gone; `sprint_event` is the shape now), and `src/alsa_raw_test.c` is a standalone scratch program nothing builds. #563 stays open for both.

`amy_midi.h` also picks up `stdbool.h`/`stdint.h` of its own — it declares `bool` and fixed-width ints and was getting them only from whatever the includer had already pulled in, which held exactly as long as every caller reached it through `amy.h`.

## A guard worth having anyway

`MIDIGetNumberOfSources()` in a process that has never made a MIDI client does not answer — it **blocks**. `midi_out()` has carried a guard against exactly this for the same reason. The port functions are the first thing a host calls, often before `amy_start()`, so without the same guard the answer to "what ports are there" is a hung process with no output and no error. Found the hard way. With no client there are no ports, which is also the truth.

## Testing

**macOS.** Audio enumeration verified live against a real desk — 9 outputs, 6 inputs, names correct and in miniaudio's order — including selecting one and rebuilding the device. The CoreMIDI half compiles clean and is wired end to end but could not be exercised in the shell it was written from, where `MIDIClientCreate` blocks outright (no GUI session); that is what turned up the guard above. **It wants a run on a machine with a controller attached before merging.**

**Linux**, in a `debian:bookworm-slim` container with `libasound2-dev`: `src/linux_midi.c` compiles clean under `-Wall -Wextra`, the whole library builds and `amy-example` links with all four `amy_midi_port_*` symbols resolving to it, and a probe run on a machine with **zero sound cards** answers 0 ports / selected -1, falls back rather than erroring on an out-of-range `select()`, and leaves an empty string for an out-of-range `name()` — no crash on the empty walk. Not yet exercised against real hardware or a real virtual port.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1327

PR #1123, opened 2026-08-19
Recheck synth[] before following one osc's reference to another

bwhitman:
Fixes a SIGSEGV on the audio thread introduced by osc-free-on-release (#1105 / #1107).

## What happens

Freeing an osc NULLs its `synth[]` slot, but nothing walks the rest of the array to clear the references **to** it. `mod_source`, `algo_source` and `chained_osc` are all stored as bare osc numbers, and there is no back-pointer to invalidate them. While `synth[osc]` stayed allocated for the whole session, a stale reference was at worst a wrong sound; now it is a null deref inside `amy_render`.

`amy_render()` skips a NULL osc at the top of its loop, so the *referrer* is safe. Each descent past that skip needs its own check — which `render_osc_wave()` already does before following `chained_osc`:

```c
if (synth[chained_osc] != NULL && synth[chained_osc]->status == SYNTH_AUDIBLE) {  // We have to recheck this since we're bypassing the skip in amy_render.
```

Seven other places did not:

| | |
|---|---|
| `envelope.c` `compute_mod_scale` | `hold_and_modify(source)` — **the reported crash** |
| `algorithms.c` `render_algo` | `synth[algo_source[op]]->role` |
| `algorithms.c` `algo_note_on` / `algo_note_off` | `algo_source[i]` |
| `amy.c` `play_delta` | the `mod_source[m]` note-on trigger loop |
| `amy.c` `play_delta` | three `chained_osc` walks (MIDI_NOTE, note-on, note-off) |
| `amy.c` `render_osc_wave` | the fall-silent `chained_osc` walk |

Each is a one-line recheck; none changes behaviour for an allocated osc.

## How it was found

A crash report from a desktop (macOS / miniaudio) build: `EXC_BAD_ACCESS` at `0xc` on `com.apple.audio.IOThread.client`. The binary was stripped, but the instruction bytes in the report identify it exactly —

```
ldr  x8, [x20]                  ; synth
ldr  x8, [x8, w0, uxtw #3]      ; synth[osc], osc = 0x76 = 118 -> NULL
ldp  s1, s2, [x8, #0xc]         ; ->midi_note, ->velocity     far = 0xc
```

`hold_and_modify+68`, reached by `amy_render → render_osc_wave → hold_and_modify → compute_mod_scale → hold_and_modify`. Worth noting that every Juno patch names osc 1 as the mod source of its other five oscs (`v0…L1Z v2…L1Z v3…L1Z v4…L1Z v5w5L1Z`), so a single Juno voice carries five references to one osc.

## Tests

Four cases added to `tests/test_osc_free_on_release.c`, one per reference kind. They build the dangling reference the way it arises in practice rather than by calling `free_osc()` directly: a bare osc names an osc that no voice owns yet, the voice allocator (which only skips oscs with `osc_to_voice` set) later hands that same osc to a voice, and the voice's release frees it.

Each case segfaults on `main` and passes here. `make ctest` green, `make qtest` 132/132 unchanged, `make check-c-api` clean.

## What this does not do

It stops the crash; it does not stop a reference going stale. A freed osc that is re-allocated for something else still leaves the referrer pointing at whatever now lives at that number — a wrong modulator rather than a crash. Actually fixing that means either clearing back-references when an osc is freed, or refusing to free an osc that is still named. Both are bigger changes than this, and neither is needed to stop the segfault.

### DIFF
--- a/src/algorithms.c
+++ b/src/algorithms.c
@@ -153,7 +153,8 @@ void note_on_mod(uint16_t osc, uint16_t algo_osc) {
 
 void algo_note_off(uint16_t osc) {
     for(uint8_t i=0;i<MAX_ALGO_OPS;i++) {
-        if(AMY_IS_SET(synth[osc]->algo_source[i])) {
+        if(AMY_IS_SET(synth[osc]->algo_source[i])
+           && synth[synth[osc]->algo_source[i]] != NULL) {
             uint16_t o = synth[osc]->algo_source[i];
             AMY_UNSET(synth[o]->note_on_clock);
             synth[o]->note_off_clock = amy_global.total_blocks * AMY_BLOCK_SIZE;
@@ -168,7 +169,8 @@ void algo_note_off(uint16_t osc) {
 void algo_note_on(uint16_t osc, float freq) {
     msynth[osc]->logfreq = logfreq_of_freq(freq);
     for(uint8_t i=0;i<MAX_ALGO_OPS;i++) {
-        if(AMY_IS_SET(synth[osc]->algo_source[i])) {
+        if(AMY_IS_SET(synth[osc]->algo_source[i])
+           && synth[synth[osc]->algo_source[i]] != NULL) {
             note_on_mod(synth[osc]->algo_source[i], osc);
         }
     }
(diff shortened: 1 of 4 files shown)


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1329

PR #1127, opened 2026-08-20
Ignore osc references that reach outside the voice being configured

dpwe:
## The bug

An osc number in a command addressed to a synth is **voice-relative**: `base_osc` is added to reach the real osc. Nothing checked the result, so a number past the end of the voice addressed — or pointed at — whatever happened to live there, which is another voice of the same synth, or another synth entirely.

```python
amy.send(synth=1, num_voices=2, oscs_per_voice=1)
amy.send(synth=2, num_voices=2, oscs_per_voice=1)
amy.send(synth=2, osc=0, freq=4)
amy.send(synth=1, osc=0, freq={'mod': 0.1}, mod_source=1)   # synth 1 has ONE osc
```

That last line used to FM synth 1 from synth 2's oscillator. The new ctest, run against the pre-change sources, puts it plainly: `FAIL osc base+1 untouched (OVERWRITTEN)`.

## The change

Every place that combines an osc number with a `base_osc` now takes the voice's osc count and checks against it, through one helper:

```c
bool osc_ref_within_voice(int rel_osc, uint16_t oscs_per_voice, const char *what);
```

- **`amy_event_to_deltas_queue(e, base_osc, oscs_per_voice, queue)`** — checks the osc the event addresses, plus each reference it carries: `chained_osc`, `mod_source[]`, `algo_source[]`. An out-of-range reference is dropped and the rest of the event still applies; an out-of-range target drops the osc-addressed part.
- **`add_deltas_to_queue_with_baseosc(d, base_osc, oscs_per_voice, queue, time)`** — the same for a stored patch replayed into a voice.
- **`oscs_per_voice == 0` means no voice context** — absolute osc numbers from the C API and the MIDI-mapping path, or a patch being *described* rather than played — and nothing is checked.

The message names the parameter and the bound (`mod_source osc 1 is outside this voice's 1 osc, ignored`). It prints once per voice, which is where the combination happens.

`EVENT_TO_DELTA_WITH_BASEOSC` is gone; its last user was `reset_osc`, whose payload is an osc number sometimes and a mask of `RESET_*` bits the rest of the time. That choice now sits at the call site, decided the way `play_delta` decides it: an osc number goes through the checked macro, a mask through `EVENT_TO_DELTA_I` untouched. Incidentally that stops something meaningless — the old macro added `base_osc` to masks too, so `reset=RESET_ALL_OSCS` on a synth arrived as `8192 + base_osc` and only worked because the mask bit survives the addition.

## The subtlety: the bound has to be read live

A patch string can **re-shape its own voice as it runs**. The drum kits open with `if3iv1in38Z`, so `patch_oscs[384]` says 1 osc until that first command executes and turns the voice into 38. My first version passed the count known before the string ran and rejected the entire rest of the patch — 14 tests failed, every drum kit silent. `oscs_in_voice_now()` re-reads the owning voice's size per command instead.

## Testing

- **`tests/test_voice_osc_range.c`** (new, in `make ctest`): what is refused (`mod_source`, `chained_osc`, `algo_source`, an addressed `osc`, a `reset` carrying an osc number — each past the voice), what is kept (the same in range; a `reset` carrying a mask), that a no-`synth` event is absolute and unbounded, that the osc a refused reference would have reached is untouched, and that a patch re-shaping its own voice is **not** fought by the bound (all 38 kit oscs configured). **5 checks fail against pre-change sources.**
- **`amy/test.py: TestModOscOOB`** — the rendering case this started from. Its reference was generated only after confirming the audio with the out-of-range `mod_source` is bit-identical (`max abs diff 0.0`) to the same patch with that parameter never sent: "ignored" really means ignored.
- `make ctest` 11/11, `make test` 133/133.

## Docs

`docs/synth.md` gains a note that osc numbers in synth-addressed commands are voice-relative, that `mod_source`/`chained_osc`/`algo_source` are too, and that anything outside the voice is refused rather than landing on a neighbour. `docs/api.md`'s `v` row points at it.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1331

PR #1129, opened 2026-08-20
Sampler features: sub-block PCM note-ons (sample_offset) and granular time-stretch/pitch-shift (fit)

bwhitman:
Two new PCM oscillator parameters that push AMY's PCM playback toward a real sampler, plus a set of `experiments/sampler/` scripts that exercise and verify them.

## Wire encoding: a `p` sub-namespace

Per @dpwe's review, these don't take top-level wire letters. 44 of the 52 single letters are already allocated in `_KW_MAP_LIST` — only `e q r t C E U Z` are unclaimed, and half of those are retired rather than genuinely free — while the sampler side is the part still growing (keygroups, drum kits, sample-granular event times).

So `p` becomes a sub-namespace, matching the existing `i` synth layer, `z` transfer layer and `ww` wave submode:

| code | keyword | meaning |
| ---- | ------- | ------- |
| `p<n>` | `preset` | selects PCM/wavetable preset `n`, exactly as before |
| `po<n>` | `sample_offset` | sub-block note-on offset |
| `pF<n>` | `fit` | time-stretch / pitch-shift target in ticks |

19 multi-char codes already exist, and all three API generators split codes as `[:-1]`/`[-1]`, so this needed no generator changes — just the map entries and a regen. `p` never swallows a sub-letter in either ordering: `po100pF96p10` and `p10po100pF96` both parse bit-exact against the kwargs path.

## `sample_offset` (`po`) — sub-block note-on placement

AMY events execute on render-block boundaries (256 samples, ~5.8 ms). `sample_offset=k` starts a PCM note-on at sample `k` *within* the block its note-on fires in, leaving the head of the block silent.

That makes slices of arbitrary length schedulable end-to-end: a slice of length `L` starting at absolute sample `S` is followed by one at block `(S+L)//256`, offset `(S+L)%256`, and the two butt-join sample-accurately. A chopped break reconstructs against its source at **-57 dB residual**, versus **+7 dB** when note-ons are quantized to blocks.

Sticky per osc like other params; send 0 to clear.

To make those joins hold, untransposed PCM playback (no `note`, no freq mods) now uses the preset's native rate exactly, instead of the `log2f`/`exp2f` roundtrip that drifted about one sample every few seconds.

## `fit` (`pF`) — granular time-stretch / pitch-shift

`fit` decouples duration from pitch, non-destructively, engaged at note-on:

- `fit=N` (N > 0): play the sample over exactly N sequencer ticks with a pitch-invariant time stretch — it tracks `tempo`, and `note` still transposes without changing duration.
- `fit=0`: pitch-shift at the original duration.
- `fit=-1`: off.

The implementation in `pcm.c` is a synchronous granular overlap-add: two Q15 Hann-windowed grains at 50% overlap (COLA), Q16 fixed-point read steps, with a WSOLA-style ±64-frame correlation search aligning each new grain's phase against the still-playing one. Without that search a pure tone lands on an ~86 Hz comb rather than the target pitch.

No FFT and no float in the render path — roughly **2.5x** the cost of plain PCM, sized for MCU targets. Looping modes (`ww`) work: the input timeline wraps at the loop marks. Streamed `disk_sample` presets are refused at note-on (the engine needs random access).

## `experiments/sampler/`

Offline, numpy + stdlib only; each demo renders a deterministic WAV or runs `--live`.

- **`slice_breaks.py`** — spectral-flux onset slicer; slice points snapped to zero crossings, slices tile the source exactly, writes WAV slices + a JSON manifest.
- **`play_sampler.py`** — five demos: `reconstruct` (the -57 dB measurement above), `hits` (16th-grid chops, each at a different sub-block offset), `loops` (breaks `fit=` to the session tempo and looped by the sequencer), `pitch` (a `fit=0` semitone ladder at constant duration).
- **`clean_breaks.py`** — normalizes a folder of mixed-format break recordings (WAV/AIFF, extensionless AIFF, Sun `.snd`, MPC2000 `.SND`) into uniform loopable breaks: mono 16-bit 44.1k, peak-normalized, each cut to exactly one verified 4-beat bar on a detected downbeat, with a manifest. Bar detection is overlap-normalized onset-envelope autocorrelation with a 4-vs-3-beat meter gate and comb-aligned beat phase; material without a detectable 4/4 grid is skipped and reported. Over a 227-file crate: 109 verified one-bar breaks (1.6–3.2 s, 75–151 bpm), 118 skipped with reasons.
- **`play_cleanbreaks.py`** — plays random cleaned breaks back to back at one fixed `--bpm`, each `fit=` stretched onto its own bar count so a mixed-tempo crate locks to a single grid, with transitions on sequencer bar boundaries.

`README.md` there carries the design write-up, verification results, and next steps (keygroup/drum-kit instrument layer, sample-granular event times).

## Compatibility and tests

Nothing changes unless the new params are used — all 132 golden tests pass, including the wire-parser fuzz. Beyond the goldens: `sample_offset=128` delays the note-on onset by exactly 128 samples, `fit=96` at 120 bpm renders ~1.0 s, and the patch string `patches.c` now serializes (`v0w7po100pF96p10Z`) round-trips through the C parser bit-exact (-240 dB). Generated Python/JS/Godot APIs were rebuilt via `make c-api` / `make godot-api`, and `docs/api.md` documents both params under a "Sampler timing and time-stretch" section.


dpwe:
Rather than using two top-level letters, how about we make `p` be a prefix for pcm-related config? So e.g. `po` for sample offset, `pf` for ticks fit.

bwhitman:
> Rather than using two top-level letters, how about we make `p` be a prefix for pcm-related config? So e.g. `po` for sample offset, `pf` for ticks fit.

Jinx, that is done

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1332

PR #1131, opened 2026-08-21
fit_search ('pS'): make the WSOLA grain-alignment search width a runtime param

bwhitman:
The `fit` stretcher's WSOLA alignment search half-width was a compile-time `#define` at 64 frames. That is only wide enough to reach half a period above about `AMY_SAMPLE_RATE/128` (~345 Hz at 44.1 kHz). Below that the aligner can't see the aligned splice at all, and locks onto a line of the comb spaced at the hop rate instead — **a 110 Hz tone stretched 2x comes out at 141 Hz.**

`fit_search` (wire `pS`) makes it a per-osc param, re-read every block so it can be swept while a note sounds.

- **unset** → the compiled-in default. Bit-identical to before; all 133 tests still at −100 dB.
- **`fit_search=0`** → search off, plain fixed-grid overlap-add. Roughly what the 90s Akai CYCLIC time stretch did, and it fails the same characteristic way: the splice error repeats at exactly the hop rate, so the artifact is a stable pitched comb rather than a diffuse warble. Wrong by tens of dB on tonal material; as an effect, it's the sampler-timestretch sound.
- **over 512** → clamped rather than refused, so a part can ask for "as wide as you can" without knowing the build's budget.
- `PCM_STRETCH_SEARCH 0` still compiles the whole search out for very small parts, and then `pS` does nothing.

### Widening the range alone wasn't enough

The 64-sample comparison window is a seventh of a period at 110 Hz — too short for the correlation to tell one lag from another, so a wider range just searched more places badly. The window now grows with the search, but the tap count stays at 64 and the stride opens up to match. Cost stays linear at `(2*pS+1)*64` MACs per hop, and the int32 accumulator keeps its headroom. The decimation aliases; acceptable because the wide settings exist for bass, where the taps still land many per period.

### Measured

`experiments/sampler/fit_search_test.py` — self-asserting, deterministic offline render, no golden files:

| tone | `pS` | in-band vs out-of-band | peak |
| --- | --- | --- | --- |
| 110 Hz | 0 | −98.9 dB | 141.4 Hz |
| 110 Hz | 64 (default) | −15.0 dB | 141.4 Hz |
| 110 Hz | 128 | +13.8 dB | 112.1 Hz |
| 110 Hz | 256 | +26.1 dB | **110.0 Hz** |
| 440 Hz | 0 | −103.4 dB | 478.6 Hz |
| 440 Hz | 64 (default) | +19.8 dB | 439.3 Hz |
| 440 Hz | 256 | +37.9 dB | 440.0 Hz |

440 Hz is already inside the default's reach, so its improvement is the longer comparison window alone. That suggests the default is leaving quality on the table for mid-range material too, but raising it quadruples the search cost for everyone, so this PR leaves it alone.

### Checks

- `make ctest`, `make test` (133 pass, err −100 dB), `make check-c-api` all clean
- builds clean with `-Wall -Wextra`, and with `PCM_STRETCH_SEARCH 0`
- wire round-trip verified: `amy.send(osc=0, fit_search=256)` → `i1v0w7p1pS256Z` → replays identically
- generated APIs rebuilt (`make c-api`, `make godot-api`)

### Follow-ups (noted in the experiment README, not done here)

- the loader could estimate the right width from the sample's lowest pitch instead of leaving it to the caller
- the search maximizes a plain dot product, which is biased toward loud candidates — invisible at ±64, but the wide settings can reach far enough to be pulled onto a transient. Normalizing by candidate energy would fix it.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1333

PR #1134, opened 2026-08-21
Distortion follow-ups: stage stacking, drive/mix modulation, per-bus stage

rt-rtos:
### 08-26 contextual shape update

Pushed the contextual shape. `J` is gone; the five `G` commands take their
scope from the event: one that names an osc shapes that osc, one that names
none shapes the bus it addresses - `bus=`, else the bus its synth is on, else
bus 0. So `synth=1, dist_clip=1` lands on the synth's bus, as you read it, and
`bus=` next to an explicit osc stays routing. `BUS_DIST_*` remain separate
delta params, since a delta names one target. At bus scope only the constant
term of `GD`/`GM` is read.

While in there: `set_event_for_osc` never emitted the `dist_*` fields (since
#1116), so `dump_state()` dropped a voice's distortion, and `num_voices=`
growth - which rebuilds every voice from a readout of voice 0 - stripped it
from the voices that had it. Both read back now. `tests/test_dist_scope.c`
covers the four scope cases and the round trips on both sides.

---

## Summary

The three follow-ons from #1116, as one PR with one commit per feature so
each stays separately reviewable and revertible. Each was built and demoed
as a prototype branch during the #1116 review (/dist-stack,/dist-coef-rail-v2,
 /dist-bus-stage); here they are rebased onto current main and adapted where they interact. 
Folding the three into one PR since they overlap heavily in the dist plumbing - can be split if requested.

#### **1. Stage stacking.** `dist_config`'s type slot becomes a stage bitmask:
`GC`/`GF`/`GH` become true per-stage toggles (GC1 no longer turns off an
enabled crusher) and state-to-wirecode collapses to 1:1 - `GH0` prints as
`GH0`. Wire and existing patch strings are unchanged; multi-stage messages
like `GC1GH6,5` round-trip verbatim. Enabled stages run as their own passes
in a fixed clip -> fold -> crush order; every pass keeps the zero-overhead
loop form, so cost is additive - roughly 70 cycles/sample/osc with all
three on ESP32-S3.

#### **2. Drive and mix on the control-coef rail.** `GD`/`GM` grow from scalars
into coef lists; `parse_coef_message` on a single value sets just the
constant term, so `GD8` means what it meant. Drive rides a log2 rail
exactly as freq and filter freq do (constant in linear drive, modulation
coefs in octaves, range 2^-4..2^4); mix follows duty (linear, clamped
0..1). Velocity into drive, EG1 into drive and mod sources into drive all
come for free from the existing machinery.

On the idle-coef cost you raised: the combine in `hold_and_modify` is
gated on `dist_stages`, so an osc with no stage enabled pays one compare
per block and no coef math. `dist_block`'s loops are untouched - drive and
mix were already hoisted above the sample loops.

#### **3. Per-bus stage.** `dist_block` over each bus's osc sum, first in the
bus FX chain, so echo and reverb take tails of the shaped signal, and the
stage responds to the whole bus level - what a mixbus saturator is for.
`J` mirrors the `G` grammar (per-stage enables via a shared parser;
`JD`/`JM` stay scalar - a bus has no per-note modulation sources to feed a
coef vector). The stage pre-clamps its input to 8x full scale as the
32x32-multiply fallback's wrap guard. This is per-bus only - no final-mix stage 
(everything on one bus gets you whole-mix distortion, pre-FX = clean tails -
arguably the better default anyway).

Bus distortion makes it much easier to commit crimes against music -
the musical ranges are narrower than per-osc (you want some dry in the mix,
modest drive, nobody needs a 16-deep fold on a full bus).
Those are aesthetic ranges, though, not safety ranges.
The engine has no insight the embedder lacks here,
so it exposes the full range and leaves taste to the caller.

## Verification

Host build clean, `make check-c-api` zero drift, ctest 12/12 including
`tests/test_dist_coefs` (pins the drive-coef octave scale positively and
negatively - a velocity coef of 2 octaves at full velocity renders
identically to a stated drive of 4, a coef of 1 does not.

Bus stage re-verified after the rebase:
enable-then-disable renders bit-identical to never-enabled,
a bus-1-directed config leaves bus 0 byte-identical,
crush engages, and stacked bus stages render distinct from single ones.


dpwe:
Looks good. 

Rather than having `JC` etc mirroring `GC` but for buses, I'm temped by making `GC` etc apply to a bus depending on the context of the event.  So:
```
amy.send(synth=0, osc=1, dist_clip=1)
```
sets clipping distortion on a particular voice oscillator, but
```
amy.send(bus=1, dist_clip=1)
```
sets it for the bus.

I think this is not ambiguous?  It means you cannot use a default value of `osc=0`, you have to explicitly specify an osc to apply dist to the osc - and I guess explicitly specify a bus to have it apply to a bus?  `amy.send(osc=1, bus=2, dist_clip=1)` is explicitly setting the bus of the osc, so it also sets the dist of the osc, not the bus that happens to be mentioned in the same event.  `amy.send(synth=1, dist_clip=1)` seems ambiguous, but we can interpret it to mean "set the distortion for the synth's bus (since you didn't specify an osc)".  This makes the processing in `amy_event_to_deltas_queue` a little complicated (I agree with keeping separate `d.param` values for `DIST_CLIP_EN` and `BUS_DIST_CLIP_EN`.

Maybe there's some gotcha to this syntax I haven't thought of?

rt-rtos:
It works, and it is consistent with how `bus=` already behaves: `bus=` with an
explicit osc routes that osc, `bus=` without one addresses the bus, so `GC`
following the same rule would be consistent. Two considerations from tracing it through `amy_event_to_deltas_queue` and `patches_event_has_voices`:

1. **`synth=1, dist_clip=1` already means something.** An osc field with
   `synth` set and `osc` unset fans out to every osc of every voice in the
   synth (`patches_event_has_voices`, the "send to all oscs in voice"
   branch) - the same as `synth=1, filter_freq=...`. Under the context rule it
   becomes the synth's bus instead, which is also whatever other synths share
   that bus; "distort synth 1"  distorting synth 2 (already true for this type of routing,
   but dist is a nastier suprise).
   `osc` keeps its 0 default for every other field in the same message and loses it for this one.

2. **Drive and mix are not the same type at the two scopes.** `GD`/`GM` are
   coef lists; `JD`/`JM` are scalars because a bus has no per-note sources. On
   one letter the bus form has to accept the coef grammar and either keep only
   the constant term or reject the rest - solvable, but a documented asymmetry
   either way.

Smaller: a message stops being decodable on its own. `GC1` needs the presence
or absence of a `v` in the same message to say where it lands, which the wire
table and anything that prints or reads wire (`sprint_event`, amy-message)
has to know about.

None of that is blocking. The way I read the tradeoff: the explicit split
buys scope you can read off the message, at the price of one more wire
letter; the context rule saves the letter and takes on the side effects
above.

I still lean towards the split. My default is a standing skepticism of opaque processes -
the same value landing in different places or meaning different things depending on context.
But this may be exactly the case where a making a considered exception is the correct tradeoff.

How heavily the letter budget weighs is unknown to me - 
you have a better grasp of how much a wire letter is worth and what else has a claim on it,
so I will leave it up to you.

If you want the contextual scoping I will add it to this PR. 

Rough outline:
`event_addresses_bus` / `event_addresses_oscs` looking at `e->osc` for the
dist fields, plus the bus-side parse of `GD`/`GM` taking only the constant
term.

dpwe:
I think my overriding principle is making the syntax as intuitive as possible.  `amy.send(bus=0, dist_clip=1)` is a syntax error if the correct keyword is `bus_dist_clip`, but to the user it's "well what did you think I meant?".  Unlike machine-to-machine communications, in which explicit distinctions are usually a good choice, I see the wire code (isomorphic to `amy.send()` arguments) as HCI  and I want to avoid telling the user "I know what you wanted was clear to you, but *technically* you should have used the `bus_dist_*` versions of the keywords for that."

rt-rtos:
Hadn't looked at it from the sender's side. It's the same worry - meaning that depends on context the message doesn't show - but yours is for the person writing it and mine for the person reading the code that resolves it, and there are far more of the former. 

Updated with new shape + some minor things I missed in #1116  (`set_event_for_osc` `dist_*` fields)

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1339

PR #1137, opened 2026-08-22
Join the second-core render with a dedicated semaphore

rt-rtos:
`esp_render_on_cores()` joins the second-core worker with a plain task
notification on the calling task's default notification index. That counter is
not private to the join: the calling task belongs to the app (with
`multithread = 0` it is whatever task runs `amy_update()`), and pacing such a
task with task notifications is idiomatic FreeRTOS - a timer ISR giving one
notification per audio block, for example. When a tick lands inside the join
window, the join returns on the tick instead of the worker: the caller runs
`amy_fill_buffer()` while the worker is still writing `fbl[1]` and combines a
half-written buffer. The worker's real "done" give then miscounts the app's
next wait, desyncing its clock. It presents as intermittent, load-dependent
garbling that only appears in multicore configs.

This replaces the caller-side join with a dedicated binary semaphore, which
nothing else can signal, and retires `amy_render_task_done_handle`. The worker
kick keeps its notification: the worker task is AMY-internal and nothing else
notifies it.

Indexed notifications (`xTaskNotifyGiveIndexed` on a non-zero index) would
also work but require `configTASK_NOTIFICATION_ARRAY_ENTRIES > 1`, which the
supported cores don't all guarantee; the semaphore is portable.

## Verification

A/B on ESP32-P4 (ESP-IDF 6.1, `multicore=1`, `multithread=0`), with the
`amy_update()` task paced by a GPTimer ISR giving one default-index task
notification per block - the pacing shape described above. To make the
overlap deterministic, the second-core render window was stretched past the
5.33 ms block period so every join spans a tick; a worker-busy flag checked
at join return counts early releases.

- Unpatched join: 23,929 early releases in a 2-minute run (~180/s,
  essentially every block), the first within 3 s of boot, with the expected
  knock-on desync as the join consumes the app's pacing ticks.
- With this patch: zero early releases; a separate 17-minute soak
  (194,000 joins) ran with no deadlock and no missed completion.

Originally observed on ESP32-S3 with the same GPTimer pacing: intermittent,
load-dependent dual-core corruption that stopped once the join no longer
shared the task's notification counter. Compile-verified on ESP32-S3
(ESP-IDF 6.0.2) as well.

### DIFF
--- a/src/i2s.c
+++ b/src/i2s.c
@@ -13,6 +13,7 @@
 
 #ifdef ESP_PLATFORM
 #include <esp_task.h>
+#include "freertos/semphr.h"
 
 ///////////////////////////////////////////////////////////////
 // ESP32, S3, P4 (maybe others)
@@ -30,7 +31,7 @@
 //   * esp_render_task (the task that runs on the second core):
 //      - Waits for notification (from main fill_audio thread)
 //      - renders half the oscs
-//      - notifies either fill_buffer (multithread) or amy_update (single thread) when done.
+//      - gives esp_render_done_sem when done, releasing whichever task ran esp_render_on_cores.
 //      - loop
 //   * esp_fill_audio_buffer_task (launched as one parallel task):
 //      - if not using I2S, waits for notification (from amy_update task)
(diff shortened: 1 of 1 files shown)


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1337

PR #1143, opened 2026-08-29
Add amy.version, and bump it + pyproject.toml on release

dpwe:
## What

Adds a `version` string to the `amy` Python package, and extends the release workflow to keep it — and `pyproject.toml`'s `version` — in lockstep with `library.properties`.

Three files now move together in the single commit the release workflow tags:

| File | Line | Consumer |
|---|---|---|
| `library.properties` | `version=1.2.160` | Arduino Library Manager; also the source this workflow reads to compute the next number |
| `amy/__init__.py` | `version = '1.2.160'` | `import amy; amy.version` |
| `pyproject.toml` | `version = "1.2.160"` | what pip reports for an installed `amy` |

## Why

There was no way to ask a running AMY which release it came from. The version lived only in `library.properties`, which a pip install doesn't ship and which doesn't survive the MicroPython freeze or the WASM build — so `amy.version` is a **literal**, not a runtime read of `library.properties`.

## Notes for a reviewer

- **`pyproject.toml` jumps `0.1.0` → `1.2.160`.** This changes what pip reports for the package. Intended, but it's a visible jump for anyone pinning `amy==0.1.0`.
- **Each `sed` is followed by a `grep` guard.** If an anchor line is ever reformatted so the `sed` silently no-ops, the release fails loudly rather than shipping a tag whose version strings lag `library.properties`.
- **Both Python anchors are now `^version = `,** one per file. `amy/__init__.py` and `pyproject.toml` each have exactly one such line today, and each `sed` names its own file — but an unaddressed `s///` rewrites every match, so if a future edit adds a second `version = ` line to either file, that `sed` would rewrite both. The `grep` catches a no-op, not an over-match.
- **No CI drift check was added.** Unlike `check-c-api`, nothing fails a PR that hand-edits one of these three without the others. The workflow overwrites all three from `$NEXT` on the next release, so drift self-heals within one merge. Happy to add a comparison step to `c-cpp.yml` if you'd rather it be enforced.

## Testing

- `python3 -c "import amy; print(amy.version)"` → `1.2.160`
- Ran the workflow's `sed` + `grep` pairs against scratch copies of both files with `NEXT=1.2.161`; both rewrote and passed their guards, the rewritten `__init__.py` still parses via `ast`, and the rewritten TOML parses via `tomllib`.
- `release.yml` parses via `yaml.safe_load`.

The workflow itself has not been exercised end to end — that only happens on the next merge to `main`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,8 +7,13 @@ name: Release
 # WHY THE BUMP GOES THROUGH A PR: main is a protected branch (a PR is required to
 # merge, and enforce_admins is on), so the "Bump version to X" change to
 # library.properties cannot be pushed straight to main — not even by an admin or
-# a PAT. This workflow therefore opens a one-line bump PR and merges it, matching
-# the manual claude/bump-* flow that produced every release to date.
+# a PAT. This workflow therefore opens a bump PR and merges it, matching the
+# manual claude/bump-* flow that produced every release to date.
+#
+# WHAT THE BUMP TOUCHES: library.properties (the Arduino version, and the source
+# this workflow reads to compute the next one), amy/__init__.py's `release`
+# string, so `import amy; amy.release` reports the tag that copy shipped in, and
+# pyproject.toml's `version`, so a pip install reports the same number.
 #
 # WHY THIS DOESN'T LOOP: the bump branch, PR, and merge are all done with the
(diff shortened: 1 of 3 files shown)


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1340

PR #1149, opened 2026-08-30
Restore real drum-kit osc counts in patch_oscs (kits 385+ play a sine after any kit switch)

bwhitman:
## Bug

Since ddafbd9 (2026-07-24, the #966 osc-table overrun fix), loading any Gamma9001 drum kit other than 384 onto a fragmented oscillator pool — the normal state once any synth was allocated after the drum synth's block, i.e. right after default boot — leaves the kit's oscillators completely unconfigured. Every drum note then plays its note-map template (`n60…`) on a default osc: a **~261 Hz sine instead of a drum**. Surfaced by a user report against the `turning-machine-4` AMYboard World sketch ("changing the drum kit plays a sine wave instead of the kick; 384 works, 385+ don't"), reproducible on AMYboard hardware and the web editor.

## Mechanism

ddafbd9 changed `patch_oscs[]` for the drum patches (258, 384–390) from their real counts to a placeholder 1, relying on the patches' leading `if3iv1in42Z` header to size the synth. But `patches_load_patch()` uses `patch_oscs` for the **initial** allocation, and `parse_patch_string_to_queue()` queues all the per-osc kit config deltas against **that** block's base osc. The in-string `inXX` realloc then runs a *different-sized* free-block search — growing from 1 to 38/42 oscs — and under fragmentation lands at a different base. The config deltas hit the old base; `dump_state` shows `i10if3iv1in42Z` followed by 42 empty `i10v0Z…` entries.

Kit 384 (38 oscs) usually escapes: switching kits frees the old kit's 38-osc hole and a 38-osc kit fits back into it at the same base. The 42-osc kits (385–390) never fit and always relocate. Bisected: good at ddafbd9^, bad at ddafbd9.

## Fix

`make_patches()` emits the real osc counts again (258→38, 384→38, 385–390→42), so both allocations run the same-size search over the same pool state and land at the same base. 258's single `patch_oscs` entry covers two build-conditional string variants (gamma808 ROM = 38 oscs, pcm_tiny ROM = 32), so it uses the larger — over-allocating 6 oscs in the non-GAMMA9001 build is harmless, under-allocating retriggers the relocation. The actual #966 fix (the block-wrap check in the allocator search) is untouched.

## Verification

- Repro (native python module, gamma bank baked in): `restart(1)` (default synths) + one extra synth, then `send(synth=10, patch=385)`, note 36 → FFT dominant **263.5 Hz** (sine) before, **53.2 Hz** (TR-909 kick) after. All kits 384–390 and 258 verified configured via `dump_state` under fragmentation.
- `make test`: 133 tests pass.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1343

PR #1152, opened 2026-09-04
BLOCK_SIZE_BITS can be chosen at compile time

bwhitman:
amy.h defined the block size unconditionally -- 256, or 128 under AMY_DAISY -- so a host could not ask for anything else without editing the header. 

Now `-DBLOCK_SIZE_BITS=5` is honoured: the default sits behind `#ifndef`, AMY_BLOCK_SIZE is derived from it, and a size that is not a power of two from 32 to 1024 is an #error. Built with nothing passed, the result is byte-identical to before -- 256, and 128 on Daisy.


dpwe:
I lean towards have AMY_BLOCK_SIZE_BITS as the thing that can be overridden, and `#define AMY_BLOCK (1<<AMY_BLOCK_SIZE_BITS)`. Doesn't particularly help with `constants.py`, but this is profound enough to be a special case. 

Having the human specify in terms of bits underlines the constraint that block size must be a power of 2. This is (probably) needed for fast indexing. (Should check the uses of `BLOCK_SIZE_BITS` in the code to see).

bwhitman:
Good point. Let me try that instead

bwhitman:
Done!

bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1348

PR #1156, opened 2026-09-09
sequencer: accumulate events on a tag instead of replacing

dpwe:
A tag used to be a slot. Sending a second `ticks="tick,period,tag"` to the same tag overwrote the first, so a tag held exactly one event and building a pattern cost one tag per note.

Events on a tag now **accumulate** -- each send adds another entry with its own `tick` and `period` -- so a tag names a whole pattern:

```python
for step, note in ((0, 36), (12, 42), (24, 38), (36, 42)):
    amy.send(osc=0, note=note, vel=1, ticks="%d,48,3" % step)   # four events, one tag
```

`ticks="0,0,<tag>"` keeps its meaning as the cancel and is now the only way to take a tag back: it drops **every** event stored under it. There is no way to remove one event from a tag while leaving its siblings -- you rebuild the tag instead. `ticks=` claims the rest of its message, so the clear has to be its own send.

No new wire code and no new keyword: the whole change rides on the existing `H` spelling.

## Implementation

The table stops being tag-indexed. It is now a slot pool with an explicit `tag` field, lowest-free allocation over `[0, max_sequences)`, and the anonymous pool unchanged past it. Two consequences worth calling out, both documented at length in the source:

**Ordering.** Same-tick firing order has always been tag order, which used to fall out free from an index-sorted active list. The list is now sorted by tag, with new entries spliced *after* equal tags, so tag order survives and cumulated events fire in the order they were added.

**Termination.** The old walk could not cycle because every link was greater than the slot holding it. Sorting by tag breaks that invariant -- a tag's slots come from wherever the pool had room -- so the tick walk is bounded at one table's worth of steps. A torn read can still make a walker skip or revisit an entry for a single tick, which is the hazard the indexed sweep always had and costs one tick of wrong events; it can no longer hang the render thread.

`max_sequencer_tags` (default 256) now caps total tagged *events* rather than just distinct tags. The footprint is unchanged, but overflow is newly reachable, so it warns and drops rather than scribbling.

`examples.c`: the two drum examples share tags 0-3 and relied on replace to start clean, so they now clear those tags before building their patterns.

## Testing

New `tests/test_sequencer_cumulate.c` (wired into `CTESTS`) covers accumulation, clearing one tag without disturbing another, the wire and C-event spellings of the clear, the bounds check that keeps a user tag out of the anonymous pool, and pool exhaustion.

- `make ctest` -- 15 binaries, all pass
- `make qtest` -- 133 render tests pass
- `make check-c-api` -- clean

One flake seen once and not reproducible: `test_sequencer_active`'s `test_cost_is_independent_of_tag` failed under parallel build load. It is a pre-existing wall-clock benchmark (`high < low * 2.0`) unrelated to this change, and it passes standalone and on every rerun -- but it will misfire on a loaded CI box.

## Downstream

`tulipcc`'s `sequencer.py` needs a matching change and has to move with the pin: its `AMYSequenceEvent.update()` re-sent `ticks=` to a tag that still held the previous version of the step, which stacks under accumulate. That is prepared, along with a redesign that gives each `AMYSequence` one tag instead of one per step (which is what ran the 256-tag space out during ordinary editing).


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1351

PR #1159, opened 2026-09-11
chorus: make chorus_max_delay bound the delay sweep

rt-rtos:
Update: 
 The test job failed because the get_synth_commands dumps include
 the chorus config and the expected strings still carried the old 320 default.
 I updated the six strings rather than keeping 320 as the default,
 since 512 is the sweep the existing patches were tuned against
 and the one the old code actually produced, so patches stay audibly identical.

---

`config_chorus()` stores `max_delay / 2` in the delay line's `fixed_delay`,
but `delay_line_in_out()` used `fixed_delay` only to place the feedback tap,
and the chorus passes `feedback_level = 0`. The modulated read tap was always
centered at half the line with a span of half the line, so `chorus_max_delay`
had no effect: the default of 320 never produced the 80..240 sample sweep its
comment describes, and every setting sounded like `max_delay = 512`.

Before the fixed-point port of `delay.c` (103337c) the runner computed the
tap as `max_delay / 2 + max_delay / 2 * mod`; the port replaced the centre
with a constant half line, and the delay line's `max_delay` field was later
removed. This restores the original formula with `fixed_delay` standing in
for `max_delay / 2`.

This centers the read tap on `fixed_delay` and scales the sweep by it, so
the delay runs `fixed_delay * (1 + mod)` = 0..`max_delay` samples.
`max_delay` is clamped to `DELAY_LINE_LEN`, the bound the `chorus_config`
comment already states.

The default moves to 512, the value the old code effectively used and the
sweep the Juno chorus settings (f1ab734) were tuned against, so existing
patches and the test references keep their sound: a saw note
rendered through the default chorus is sample-identical before and after.
The comments and generated constants follow the new default.

Verification: an impulse through `delay_line_in_out()` at constant `mod_in`,
fixed point, centroid of the delayed copy in samples.

| max_delay | mod -0.5 | mod 0 | mod +0.5 | before (any max_delay) |
|---|---|---|---|---|
| 320 | 80 | 160 | 240 | 128 / 256 / 384 |
| 128 | 32 | 64 | 96 | 128 / 256 / 384 |
| 512 | 128 | 256 | 384 | 128 / 256 / 384 |

The channel driven with `mod_scale = -1` mirrors the sweep.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1360

PR #1163, opened 2026-09-15
filters: reset the filter state when a FILTER_TYPE delta changes the type

rt-rtos:
Changing `filter_type` on an oscillator that is already sounding hands the new
kernel the previous kernel's `filter_delay`, and the kernels store different
things there. `dsps_biquad_f32_ansi_split_fb` keeps raw input history in
`w[0..1]` (and a block-floating-point shifted state on `!AMY_HAS_MUL64`
targets); the LPF24 kernel keeps input pre-multiplied by `coef[0]` in six
words; the phaser keeps an allpass chain. Read as LPF24 state, raw history is
about `1/b0` too large, some 5000x at 200 Hz / Q 4, and rings both resonant
stages: a burst at the cutoff frequency tens of dB above the note, clipping at
normal velocities and decaying over 40-50 ms. Every switch into `FILTER_LPF24`
does it, as does switching out of `FILTER_PHASER` into any biquad; switches
between the 12 dB types share a convention and are clean.

`play_delta` now calls `reset_filter` when a `FILTER_TYPE` delta carries a
different type than the osc holds. `filter_type` has no other writer, so the
check runs per event, not per block, and needs no new state. `reset_filter`
zeroes the whole `filter_delay` array and `last_filt_norm_bits`, which covers
both kernel variants. The old filter's tail is dropped at the change; it
measured no louder than the destination filter's own onset. The two
commented-out `reset_filter` calls in `amy.c` (note-on and voice-stop) are
untouched; they address a different moment.

### Audio + Spectrogram

lpf24_switch_spectrogram

**Before** (volume/speaker hazard warning)

[lpf24_switch_before.wav](https://github.com/user-attachments/files/32248807/lpf24_switch_before.wav)

**After**

[lpf24_switch_after.wav](https://github.com/user-attachments/files/32248810/lpf24_switch_after.wav)

**Before then After**

https://github.com/user-attachments/assets/a704b5bf-99e4-4f9c-8f0a-9e12f6adc6e6

Host build, fixed-point (`AMY_USE_FIXEDPOINT`, `AMY_HAS_MUL64`), saw 60 Hz,
cutoff 200 Hz, Q 4. Per-block peak, full scale 1.

| switch | velocity | before | after | destination cold start |
|---|---|---|---|---|
| LPF -> LPF24 | 0.05 | 0.508 | 0.013 | 0.011 |
| LPF -> LPF24 | 1.0 | 1.0 for 6 blocks, then 0.69 0.42 ... | 0.26 | 0.26 |
| HPF / BPF / NOTCH -> LPF24 | 0.05 | 0.509 | 0.013 | 0.011 |
| PHASER -> LPF24 | 0.05 | 0.719 | 0.013 | 0.011 |
| PHASER -> HPF | 0.05 | 0.040 | 0.007 | 0.006 |
| LPF24 -> LPF, LPF -> HPF / BPF / NOTCH | 0.05 | clean | clean | |

`tests/test_filter_type_switch.c` (added to `CTESTS`) runs every ordered pair
of the six types and checks the loudest block after the switch against twice
the destination's cold-start peak over the same window. The factor of two
sits between the fixed behaviour (within 1.2x of cold start in every pair)
and the smallest burst the change removes (5x, PHASER -> LPF). It fails in
ten pairs on main and passes with this change. Only this test was run, not
the full suite.

## Notes

- Second commit, comment hygiene only: `allpass1_chain` is the one-state transposed
  allpass with two multiplies per stage, not a one-multiply form, and its
  memories are `w[]`, as the parameter is named. Independent of the fix;
  drop it if unwanted.
- Continuity instead of a reset is possible: right-shift the state by
  `last_filt_norm_bits`, then scale `w[0..1]` by `coef[0]` on entry to LPF24.
  Not done; the reset is simpler and the dropped tail is inaudible in the
  measurements above.
- Separate observation, not changed: `dsps_biquad_gen_lpf_f32` stores `-b/a0`
  while the other generators store `+b/a0`. The 12 dB kernel applies it once
  and the LPF24 kernel twice, so 12 dB LPF output is sign-inverted relative to
  the other types.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1359

PR #1164, opened 2026-09-15
esp32: let the host set the render and fill task priorities

bwhitman:
`AMY_RENDER_TASK_PRIORITY` and `AMY_FILL_BUFFER_TASK_PRIORITY` are defined unconditionally at `ESP_TASK_PRIO_MAX - 1`, the FreeRTOS ceiling, so a host cannot put anything above the render tasks.

A host with a USB audio class device has to: TinyUSB arms each isochronous IN packet from its own task once a millisecond, and a render task at the ceiling holds that task off for the length of a block whenever a chord is playing. Measured on an ESP32-P4 with 16 DX7 voices (render load ~0.7): the packet loads came late 4,123 times in 17 s and a third of the capture stream was lost at the host. With amy's tasks one below the ceiling and the USB task at it, the loads are never late under any load amy can render, and the render's own ceiling did not move.

This wraps the four defines in `#ifndef` so a host can define its own before including `amy.h`. The defaults are unchanged for everyone who does not.


bwhitman:

This merge was pinned into tulipcc for full-system CI: https://github.com/shorepine/tulipcc/pull/1358

